PR #18, opened 2024-04-07
added fixes while initial setup(Thanks Giorgos)

ilakshani:
## Description

Fixed some points while trying to initial project set up (with support of Giorgos)

### DIFF
--- a/apps/server/package.json
+++ b/apps/server/package.json
@@ -7,7 +7,7 @@
   "main": "index.js",
   "type": "module",
   "scripts": {
-    "cross-env": "NODE_ENV=development pnpm --filter=@repo/database db:migrate:deploy && tsx watch --env-file=.env src/index.ts",
+    "dev": "cross-env NODE_ENV=development pnpm --filter=@repo/database db:migrate:deploy && tsx watch --env-file=.env src/index.ts",
     "build": "tsc",
     "start": "node --conditions=javascript --import=extensionless/register dist/index.js",
     "lint": "eslint . --max-warnings 0",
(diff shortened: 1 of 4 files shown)


apps/server/package.json
```diff
@@ -7,7 +7,7 @@
   "main": "index.js",
   "type": "module",
   "scripts": {
-    "dev": "NODE_ENV=development pnpm --filter=@repo/database db:migrate:deploy && tsx watch --env-file=.env src/index.ts",
+    "cross-env": "NODE_ENV=development pnpm --filter=@repo/database db:migrate:deploy && tsx watch --env-file=.env src/index.ts",
```
trixobird, line 10:
```suggestion
    "dev": "cross-env NODE_ENV=development pnpm --filter=@repo/database db:migrate:deploy && tsx watch --env-file=.env src/index.ts",
```
I think this is the correct amendment, no? If you agree we also need to install cross-env in apps/server and add these changes on this PR

trixobird (commented):
I think we should add on the root readme.md that you need to add an empty .env file in apps/server

packages/database/package.json
```diff
@@ -13,7 +13,7 @@
   },
   "scripts": {
     "build": "tsc",
-    "dev": "pnpm generate:watch",
+    "dev": "pnpm generate",
```
trixobird, line 16: can you undo this change? I think during the conflict we should have clicked the other options (not accept incoming :P )

trixobird (approved):

PR #53, opened 2024-05-10
Initial grouped by filters

elevchyt:
Default groupedBy filters for /insights.

### DIFF
--- a/apps/web/src/app/[locale]/(logged-in)/aside.tsx
+++ b/apps/web/src/app/[locale]/(logged-in)/aside.tsx
@@ -46,8 +46,10 @@ export function Aside({ children }: { children: React.ReactNode }): React.ReactN
   }, [pathname, router]);
 
   useEffect(() => {
-    setInitialGroupedByFilters();
-  }, [setInitialGroupedByFilters]);
+    if (pathname === '/insights') {
+      setInitialGroupedByFilters();
+    }
+  }, [pathname, setInitialGroupedByFilters]);
 
   return (
     <div className={cx('bg-menu-bg', minimize ? 'lg:w-10' : 'lg:w-1/5')}>
--- a/apps/web/src/app/[locale]/(logged-in)/aside.tsx
+++ b/apps/web/src/app/[locale]/(logged-in)/aside.tsx
@@ -4,7 +4,7 @@ import { Layers, MenuIcon, PanelLeftClose, PanelLeftOpen, Settings, X } from 'lu
 import { useCallback, useEffect, useState } from 'react';
 import { cx } from '@repo/ui/tailwind-utils';
 import Link from 'next/link';
-import { usePathname, useRouter } from 'next/navigation';
+import { usePathname, useRouter, useSearchParams } from 'next/navigation';
 import { SignOutBtn } from '@/components/login/sign-out-btn';
 import { ActiveLink, type LinkType } from '@/components/home/active-link';
 import { LogoFull } from '@/app/[locale]/(logged-in)/logo-full';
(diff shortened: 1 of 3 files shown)


apps/web/src/app/[locale]/(logged-in)/aside.tsx
```diff
@@ -1,22 +1,57 @@
...
+        key: 'groupedBy',
+        value: InsightsColumnsGroupBy.position,
+      },
+    ];
+    const newURL = createURLWithQueryParams(pathname, initialParams);
+    router.replace(newURL);
+  }, [pathname, router]);
+
+  useEffect(() => {
+    if (pathname === '/insights') {
```
Comment (trixobird) [inline apps/web/src/app/[locale]/(logged-in)/aside.tsx:50]:
```suggestion
     if (pathname.endsWith('/insights') && searchParams.size === 0) {
```
locale break the `===` also we want to make sure that it does not have any search params.
The problem here is that if for any reason the user removes all searchparams it will reset. We may add a custom `dummy=true` searchparam, but I don't like it a lot

apps/web/src/app/[locale]/(logged-in)/insights/query-string-util.ts
```diff
@@ -9,6 +9,11 @@ import {
 
 export type OrderType = 'asc' | 'desc';
 
+export type QueryParamsType = {
+  key: string;
```
Comment (trixobird) [inline apps/web/src/app/[locale]/(logged-in)/insights/query-string-util.ts:13]:
```suggestion
    key: SearchParamsKeys;
```
I like narrow typing, but maybe I this could create issues down the line?

apps/web/src/app/[locale]/(logged-in)/aside.tsx
Comment (elevchyt) [reply apps/web/src/app/[locale]/(logged-in)/aside.tsx:50]: Nice catch!

apps/web/src/app/[locale]/(logged-in)/insights/query-string-util.ts
Comment (elevchyt) [reply apps/web/src/app/[locale]/(logged-in)/insights/query-string-util.ts:13]: I think the URL will be used for more than searching, so I think we should keep it as a string for now.

apps/web/src/app/[locale]/(logged-in)/aside.tsx
Comment (trixobird) [reply apps/web/src/app/[locale]/(logged-in)/aside.tsx]: don't forget the `endsWith` if someone is using locale just `===` does not work.

Regarding the not reset to defaults on empty, I was thinking that we can add in `query-string-util.ts` after line 64
```
        if (params.getAll(key).length === 0) {
          params.set('g', 'none');
        }
```

Comment (trixobird) [reply apps/web/src/app/[locale]/(logged-in)/aside.tsx]: or something similar

apps/web/src/components/filters/grouped-checkbox.tsx
```diff
@@ -21,11 +27,22 @@ export default function GroupedCheckbox<T extends string>({
...
+  const checked = isParamInSearchParams(searchParams, groupKey, groupByColumn);
 
   function onChange(event: ChangeEvent<HTMLInputElement>): void {
     startTransition(() => {
-      router.replace(`${pathname}/?${createQueryString(groupByColumn, event.target.checked)}`);
+      // If the upcoming URL after the checkbox is pressed is about to be empty, add the groupedBy=none parameter
+      let upcomingURL = `${pathname}/?${createQueryString(groupByColumn, event.target.checked)}`;
+      if (upcomingURL.endsWith('/insights/?')) {
+        const noneParams = userActionOverrideParams;
+        const noneURL = createURLWithQueryParams(pathname, [noneParams]);
```
trixobird, line 38:
```suggestion
        const noneURL = createURLWithQueryParams(pathname, [userActionOverrideParams]);
```

trixobird (approved):
thank you for work Lefteris!

PR #56, opened 2024-05-12
Use meta instead of facebook in settings

trixobird:
## Description

Since meta has multiple publishers, with one of them being fb, let's use meta branding


apps/web/src/util/types.ts
```diff
@@ -4,6 +4,6 @@ export type UnwrapArray<T> = T extends (infer R)[] ? R : never;
 
 export const integrationTypeMap = new Map<IntegrationType, { image: string; name: string }>([
   [IntegrationType.TIKTOK, { name: 'TikTok', image: '/integrations/tiktok-logo-icon.svg' }],
-  [IntegrationType.FACEBOOK, { name: 'Facebook', image: '/integrations/facebook-logo-icon.svg' }],
+  [IntegrationType.FACEBOOK, { name: 'Meta', image: '/integrations/meta-logo-icon.svg' }],
```
elevchyt, line 7:
```suggestion
[IntegrationType.META, { name: 'Meta', image: '/integrations/meta-logo-icon.svg' }],
```

I don't think this is being used anywhere else, so it can be changed to META, too.

trixobird: this `IntegrationType` originated from a generated file, that ulitmately is traced back on the `IntegrationTypeEnum` in the database. I was a bit lazy and decided not to change it at the beginning, but your comment indicated that we still need to enforce DDD as much as possible

elevchyt (approved):

PR #80, opened 2024-05-29
introduce google login

trixobird:
## Description

providers were created to be agnostic. Adding a new provider should be a considerable easier effort


elevchyt:
@trixobird Looks good, a new .env param seems to have been added, what is its value supposed to be?

Referring to **NEXT_PUBLIC_ENDPOINT**.

trixobird:
I think this is covered by https://github.com/adsviewer/turboviewer/blob/e04a42d76a4dda8c8a2c34f66285f9de82aa7195/apps/web/src/env.mjs#L26

`NEXT_PUBLIC_ENDPOINT: z.string().min(1).default('http://localhost:3000'),` that is giving it its default value. This same trick applies for other env variables in multiple packages/apps

Issue #152, opened 2024-06-18
P2: Add data_last_fetched to integrations screen


dennislobster:
In the integrations screen, for every datasource add:

- [ ] how long ago data was fetched (i.e. 6 hrs ago)
- [ ] how large the total dataset is (i.e. 1000 advertisements) -> calculate as distinct count

### Backend Integration done

Now integrations endpoint can be filtered also by status. Additionally it retuns lastSyncedAt and in the AdAccount we have adCount

example query
```
query integrations {
  integrations {
    lastSyncedAt
    adAccounts {
      adCount
    }
  }
}
```

trixobird:
@dennislobster, let's say that an integration has this break down:
- adAccountA
  - adA with 10 insights
  - adB with 20 insights
- adAccountB
  - adC with 30 insights

What would be the number to show? Keep in mind that this answer maybe different(?) for agency and the underlying company

dennislobster:
Good question @trixobird - 

It should always show the distinct number of ad_ids for a given integration (across ad accounts the organization has access to).

So if multiple accounts, maybe show: (or roll into 1 number)
Account X: 120 ads
Account Y: 50 ads
Account ...

PR #162, opened 2024-06-26
adCount & latest sync date in ad accounts

elevchyt:
Implements the FE part of https://github.com/adsviewer/turboviewer/issues/152


apps/web/messages/en.json
```diff
@@ -32,7 +32,9 @@
       "device": "Device",
       "date": "Date",
       "publisher": "Publisher",
-      "position": "Position"
+      "position": "Position",
+      "syncedOn": "Synced on",
```
trixobird, line 36:
```suggestion
      "syncedAt": "Synced on",
```

apps/web/src/app/(authenticated)/insights/components/group-filters.tsx
```diff
@@ -39,17 +42,21 @@ export default function GroupFilters(): ReactNode {
       const integrations = res.integrations;
       let adAccounts: MultiSelectDataType[] = [];
       for (const integration of integrations) {
+        // TODO: Discuss which integration's last synced date has priority.
```
trixobird, line 45: let's keep it as it right now (you may even remove the TODO if you want)

trixobird (approved):
I have good news and bad news.

The good news are great job :) The not so bad news is that the issue didn't ask what you implemented :) it says "Add data_last_fetched to _integrations screen_" You put this on the drop down :)

I am approving this PR (there is a small type on translation), we can open another one of the other functionality

apps/web/messages/en.json
elevchyt: Ah yes, my bad.
The terminology is a bit confusing on your discussion with Dennis in that issue so by viewing the changes in graphql I assumed the last sync date is also on the adAccount since integrations are fetched there, too.

elevchyt:
Take a look again, I did it on the integrations page & removed it from the accounts' dropdown.

apps/web/messages/en.json
trixobird: I agree, I even thought that _I_ was on the wrong when I first saw them. That said I liked your changes even better than these Dennis suggested, can we keep them maybe?

trixobird (approved):

PR #173, opened 2024-06-30
110 features and fixes

elevchyt:
various fixes & improvements


apps/web/src/app/(authenticated)/insights/components/insight-card.tsx
```diff
@@ -35,50 +33,21 @@ interface Datapoint {
...
-      case 'unknown':
-        setRank({
-          label: '-',
-          color: 'gray',
-        });
-        break;
-      default:
-        break;
+    // If the latest CPM is lower than the previous CPM, rank is set to CAUTION
+    if (props.datapoints[props.datapoints.length - 1].cpm < props.datapoints[props.datapoints.length - 2].cpm) {
```
trixobird, line 44:
```suggestion
    if (props.datapoints[props.datapoints.length - 1].cpm > props.datapoints[props.datapoints.length - 2].cpm) {
```
CPM is cost per thousands, which mean that less is better

apps/web/src/app/(authenticated)/insights/layout.tsx
```diff
@@ -0,0 +1,14 @@
+import React, { Suspense } from 'react';
+import { Title } from '@mantine/core';
+import { useTranslations } from 'next-intl';
+import LoaderCentered from '@/components/misc/loader-centered';
+
+export default function RootLayout({ children }: { children: React.ReactNode }): React.ReactNode {
```
trixobird, line 6:
```suggestion
export default function InsightsLayout({ children }: { children: React.ReactNode }): React.ReactNode {
```
not sure if we should name the Layout differently, maybe too much hussle

apps/web/src/app/(authenticated)/integrations/components/integrations-grid.tsx
```diff
@@ -69,7 +72,16 @@ export default function IntegrationsGrid(props: IntegrationProps): ReactNode {
     setTimeout(() => {
       setIsMounted(true); // Set isMounted to true after a delay to allow the transition to play
     }, 0);
-  }, [updateIntegrationsData]);
+
+    // Show error notification
+    const error = searchParams.get('error');
```
trixobird, line 77: I was thinking that we can have the error component outside everything (maybe in the authenticated group, but maybe even outside), like in the [web-old](https://github.com/adsviewer/turboviewer/blob/396fdb4989235ad14a68c0a3196c7185a3db7add/apps/web-old/src/app/%5Blocale%5D/layout.tsx#L30) 

This way we can handle any type of errors maybe? Just a thought

trixobird (commented):
look great :) 

one thing that I'ld like to see, happy to create new issue (although I think it is small enough), is that the ColorSchemePicker option is not set if the selection is Auto (set it to Auto, close dropdown, open it again). Also how would you feel to set the default option to Auto?

apps/web/src/app/(authenticated)/insights/layout.tsx
elevchyt: Yeah probably not right now

apps/web/src/app/(authenticated)/integrations/components/integrations-grid.tsx
elevchyt: At the time of making this I wasn't aware that handling errors via the url is a standard practice in our project, so yes it has been generalized now to work in every view.

apps/web/src/components/misc/notifications-handler.tsx
```diff
@@ -0,0 +1,25 @@
+'use client';
+
+import { useSearchParams } from 'next/navigation';
+import { Notifications, notifications } from '@mantine/notifications';
+import { useEffect, type ReactNode } from 'react';
+import { useTranslations } from 'next-intl';
+
+export default function IntegrationsGrid(): ReactNode {
```
trixobird, line 8:
```suggestion
export default function NotificationsHandler(): ReactNode {
```

trixobird (approved):

apps/web/src/components/misc/notifications-handler.tsx
elevchyt: oops

PR #187, opened 2024-07-03
176 iframes in insights

elevchyt:
Adds support for viewing the preview of each ad (limited to grouped by Ad ID).
Testing from @trixobird is required because I had the problem with the slow query times when grouping by Ad ID.


apps/web/src/app/(authenticated)/insights/components/order-filters.tsx
```diff
@@ -42,71 +46,117 @@ export default function OrderFilters(): React.ReactNode {
...
+      router.replace(newURL);
+    });
   };
 
   return (
-    <Flex w="100%" mb="lg" align="center" wrap="wrap">
-      {/* Filters */}
-      <Flex align="center" mr="sm" my="md">
+    <Flex w="100%" mb="lg" wrap="wrap" direction="column">
+      {/* FIlters */}
```
trixobird, line 88:
```suggestion
      {/* Filters */}
```

trixobird (commented):
Code looks fine, some generic comments:

- There is a flicker when refreshing the insights page
- Clicking on the settings upper right re-renders the screen and executes again the query, that is why it is slow
- Changing the language/theme also does the same thing

elevchyt:
I removed the overlay and created a separate issue to troubleshoot why neighbor components are forcing each other to re-render on insights.
https://github.com/adsviewer/turboviewer/issues/200

trixobird (approved):

trixobird:
> I removed the overlay and created a separate issue to troubleshoot why neighbor components are forcing each other to re-render on insights. #200

I would prefer to have the flickering issue that was introduces by this PR fixed before merging it, but I'll leave this decision to you :)

PR #212, opened 2024-07-09
155 multiple orgs fe

elevchyt:
Users can now create, switch & delete organizations based on their user roles.

I'm aware of some hydration errors occurring when refreshing the organization page. These errors are only visual, the page's functionality works as intended.
However, it would be very beneficial if we could find some explanation as to why this happens. Up to now I haven't found anything related to this error that is not of the nature of wrong markdown (aka mistakes in the structure of the html).
This isn't the case with any of the pages that we encounter this error in our project.


apps/web/src/middleware.ts
```diff
@@ -48,19 +49,27 @@ export const config = {
 export async function middleware(request: NextRequest): Promise<NextResponse> {
   const token = request.cookies.get(TOKEN_KEY)?.value;
   const refreshToken = request.cookies.get(REFRESH_TOKEN_KEY)?.value;
+  let tokenData: AJwtPayload | undefined;
+  if (token) {
+    try {
+      await jwtVerify(token, new TextEncoder().encode(env.AUTH_SECRET));
```
trixobird, line 55: <3

trixobird (commented):
Some functional issues that I encoutered/suggestions:

- When an organization is created maybe it is better to automatically switch the user to that organization and stop showing the popup (I think the code tries to do it, but it is not happening)

- When I create the organization (from a non-organization setting) I don’t see any dropdown, even after I refresh. The user info bottom left is spinning for ever. Even when I sign out and sign in again, I don’t see any dropdown.

elevchyt:
> Some functional issues that I encoutered/suggestions:
> 
> * When an organization is created maybe it is better to automatically switch the user to that organization and stop showing the popup (I think the code tries to do it, but it is not happening)
> * When I create the organization (from a non-organization setting) I don’t see any dropdown, even after I refresh. The user info bottom left is spinning for ever. Even when I sign out and sign in again, I don’t see any dropdown.

I created a video that demonstrates the results from my testing.
For some reason, requesting insights right after creating the organization _(and especially before fetching user data from the me() endpoint for the first time)_, fails the request and responds with 500 after 5 mins.

https://streamable.com/04mg9u

trixobird:
> > Some functional issues that I encoutered/suggestions:
> > 
> > * When an organization is created maybe it is better to automatically switch the user to that organization and stop showing the popup (I think the code tries to do it, but it is not happening)
> > * When I create the organization (from a non-organization setting) I don’t see any dropdown, even after I refresh. The user info bottom left is spinning for ever. Even when I sign out and sign in again, I don’t see any dropdown.
> 
> I created a video that demonstrates the results from my testing. For some reason, requesting insights right after creating the organization _(and especially before fetching user data from the me() endpoint for the first time)_, fails the request and responds with 500 after 5 mins (probably the default timeout for pending requests).
> 
> https://streamable.com/04mg9u

Played around a bit. It seems that there is no request on the insights endpoint on the backend during your scenario (yes there is one after 5'). What is more there is no request on the me endpoint which I think it creates the issue. I put a breakbpoint and for some reason the `userDetails` have their initial (empty) values. This kind of makes sense since the useEffect [here](https://github.com/adsviewer/turboviewer/blob/b84c277b99db172829a734c96f9d030f101d9a1d/apps/web/src/components/user-button/user-button.tsx#L17) has setUserDetails in its deps so it won't trigger.

Happy to talk when you are online

apps/web/src/components/buttons/create-organization-button.tsx
```diff
@@ -0,0 +1,101 @@
...
+            logger.error(error);
+          });
+      })
+      .catch((error: unknown) => {
+        logger.error(error);
+      })
+      .finally(() => {
+        setIsPending(false);
+        form.reset();
+      });
```
trixobird, line 57:
```suggestion
    void createOrganization(values).then((res) => {
      setIsPending(false);
      form.reset();
      if (isAError(res)) {
        logger.error(res.message);
        return new Error(res.message);
      }
      window.location.reload();
    });
```
I think it makes more sense to move these steps on the server side, for example the `createOrganization` could end up being:
```
export async function createOrganization(values: CreateOrganizationMutationVariables): Promise<AError | undefined> {
  const createOrgResp = await handleUrqlRequest(urqlClientSdk().createOrganization(values));
  if (!createOrgResp.success) {
    return new AError(createOrgResp.error);
  }
  const switchOrgResp = await switchOrganization({ organizationId: createOrgResp.data.createOrganization.id });
  if (!switchOrgResp.success) {
    return new AError(switchOrgResp.error);
  }
  await changeJWT(switchOrgResp.data.switchOrganization.token, switchOrgResp.data.switchOrganization.refreshToken);
  return undefined;
}
```

elevchyt: Now that both are on the server side, upon creating an organization and changing the JWT token to the new token, we get the following error regarding JWS. 
This seems to sign out the user upon refresh, even if the new token has been set correctly in the cookies.
@trixobird 

```
web:dev:  [JWSSignatureVerificationFailed: signature verification failed] {
web:dev:   code: 'ERR_JWS_SIGNATURE_VERIFICATION_FAILED',
web:dev:   name: 'JWSSignatureVerificationFailed',
web:dev:   message: 'signature verification failed'
web:dev: }
```

apps/web/src/middleware.ts
```diff
@@ -100,14 +110,30 @@ export async function middleware(request: NextRequest): Promise<NextResponse> {
...
+    request.nextUrl.pathname !== '/sign-out' &&
+    request.nextUrl.pathname !== defaultMissingOrgURL
+  ) {
+    const redirectUrl = new URL(defaultMissingOrgURL, request.url);
+    return NextResponse.redirect(redirectUrl);
+  }
+
+  // If user has a current org and tries to visit the missing org screen, redirect to insights
+  if (request.nextUrl.pathname === defaultMissingOrgURL && tokenData?.organizationId) {
+    logger.info('VALID BRO');
```
trixobird, line 133:
```suggestion
```

trixobird (commented):
I think we are closing up this giagantic PR, kudos :)

apps/web/src/app/(authenticated)/organization/actions.ts
```diff
@@ -1,14 +1,63 @@
...
+): Promise<UrqlResult<DeleteOrganizationMutation, string>> {
+  return await handleUrqlRequest(urqlClientSdk().deleteOrganization(values));
+}
+
+export async function createAndSwitchOrganization(values: CreateOrganizationMutationVariables): Promise<boolean> {
+  try {
+    const res = await createOrganization(values);
+
+    if (!res.success) {
+      return false;
```
trixobird, line 47: let's log the error (applicable couple of lines down as well)

trixobird (approved):

PR #215, opened 2024-07-10
integrations access rights

elevchyt:
Limits integration actions to organization admins only.


apps/web/src/app/(authenticated)/integrations/components/integration-card.tsx
```diff
@@ -55,20 +60,25 @@ export default function IntegrationCard(props: IntegrationProps): ReactNode {
...
         return (
-          <Button
-            mt="lg"
-            onClick={() => {
-              handleConnect();
-            }}
-          >
-            {t('connect')}
-          </Button>
+          <Tooltip label={tGeneric('accessOrgAdmin')} disabled={isOrgAdmin(userDetails.allRoles)}>
```
trixobird, line 63: now that the adAccount - organizations is in main (you have to rebase), we need one extra check here. The organization needs to be the root (parentId === null)

apps/web/src/app/(authenticated)/integrations/components/integration-card.tsx
```diff
@@ -55,20 +60,25 @@ export default function IntegrationCard(props: IntegrationProps): ReactNode {
...
+              {t('connect')}
+            </Button>
+          </Tooltip>
         );
       }
       return (
-        <Button mt="lg" color={theme.colors.red[7]} onClick={open}>
-          {t('revoke')}
-        </Button>
+        <Tooltip label={tGeneric('accessOrgAdmin')} disabled={isOrgAdmin(userDetails.allRoles)}>
```
trixobird, line 77: save as above

trixobird (approved):

PR #269, opened 2024-07-22
243 migrate insights to use jotai

elevchyt:
Insights now use Jotai to better handle their state changes independent of the URL, while still utilizing the URL for filters.
As a result, changing filters is more responsive to user input, since they reset the fetched insights to empty and show a loading state before finishing the request of the new insights.

This will also improve much more when we look into integrating the dedicated animations library.

### DIFF
--- a/apps/web/src/app/(authenticated)/insights/components/page-controls.tsx
+++ b/apps/web/src/app/(authenticated)/insights/components/page-controls.tsx
@@ -5,17 +5,16 @@ import { IconArrowLeft, IconArrowRight } from '@tabler/icons-react';
 import { useTranslations } from 'next-intl';
 import { usePathname, useRouter, useSearchParams } from 'next/navigation';
 import React from 'react';
+import { useAtomValue } from 'jotai/index';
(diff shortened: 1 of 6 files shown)


apps/web/src/app/(authenticated)/insights/components/page-controls.tsx
```diff
@@ -5,17 +5,16 @@ import { IconArrowLeft, IconArrowRight } from '@tabler/icons-react';
...
-export default function PageControls(props: PropsType): React.ReactNode {
+export default function PageControls(): React.ReactNode {
   const t = useTranslations('insights');
   const pathname = usePathname();
   const searchParams = useSearchParams();
   const router = useRouter();
+  const hasNextInsightsPage = useAtomValue(hasNextInsightsPageAtom);
 
   const handlePageChange = (type: 'next' | 'prev'): void => {
     if (type === 'next') {
```
trixobird, line 20: just making sure that we don't need to invoke resetInsights here, since we are invoking addOrReplaceURLParams

trixobird (approved):
first of all great initiative. This greatly improves the experience of the user. Happy to endorse any other such ideas, like the animations

PR #301, opened 2024-07-30
247 preview card enhancements

elevchyt:
Insight cards visual and practical enhancements.
Also added more tooltips and ability to copy important information on click.


apps/web/src/app/(authenticated)/insights/components/insight-card.tsx
```diff
@@ -42,12 +48,27 @@ interface Datapoint {
...
   const t = useTranslations('insights');
+  const tGeneric = useTranslations('generic');
+  const searchParams = useSearchParams();
   const [rank, setRank] = useState<RankType>({
     label: 'GOOD',
     color: 'green',
   });
   const [datapoints, setDatapoints] = useState<Datapoint[]>([]);
 
+  const copyText = async (text: string): Promise<void> => {
```
trixobird, line 59: <3

trixobird (approved):
looks great, can you switch the logic, so the default view is the ad preview and not the charts?

PR #314, opened 2024-08-05
add userId and orgId in sentry

trixobird:
## Description

This adds userId and organizationId in sentry, in both backend and frontend. I have used a new library `cookies-next` to read cookies from the client. I am assuming that there is a better way :)


elevchyt (approved):
Nice!

PR #333, opened 2024-08-08
302 members management

elevchyt:
The core features of managing members in an organization.
Includes inviting, changing roles and removing them.


trixobird:
it looks like the last name input field shows email and not last name. Which is wierd, since if this does not happen if the email field is not pre populated
image

trixobird (commented):
in the login-providers-links.tsx you need to pass the new optional inviteHash value, in order to have the social login work

apps/web/src/app/(authenticated)/organization/components/invite-users-button.tsx
```diff
@@ -0,0 +1,165 @@
...
+                notifications.show({
+                  title: tGeneric('error'),
+                  message: errorMessage,
+                  color: 'red',
+                });
+              }
+            }
+          }
+        } else {
+          refreshMembers();
```
trixobird, line 103: not if we need this extra query, the res in line 88 should contain everything to update the users table

elevchyt: We actually need this in order to show the invited users in the grid after submitting the invitations.
You can see this in action by commenting refreshMembers() right there.

elevchyt:
> in the login-providers-links.tsx you need to pass the new optional inviteHash value, in order to have the social login work

login-providers-links.tsx seems to be a component of the web-old. Also, shouldn't the login providers be using the same sign up logic as the regular sign up?

apps/web/src/app/(authenticated)/organization/components/invite-users-button.tsx
```diff
@@ -0,0 +1,172 @@
...
+      })
+      .catch((err: unknown) => {
+        logger.error(err);
+      });
+  };
+
+  const performUserInvites = (): void => {
+    setIsPending(true);
+
+    void inviteUsers({ emails, role: DEFAULT_ROLE })
```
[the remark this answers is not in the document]
elevchyt: Sure, it's just going to be a bit wonky if you're inviting multiple members.
At the moment, the use can change each invited user's role after the invitations are successful.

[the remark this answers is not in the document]
elevchyt: Done, also made it so that Administrator role must be set explicitly after invite. The user can set initial role as Member or Operator.

apps/web/src/app/(authenticated)/organization/components/invite-users-button.tsx
```diff
@@ -0,0 +1,196 @@
...
+  const tGeneric = useTranslations('generic');
+  const theme = useMantineTheme();
+  const [opened, { open, close }] = useDisclosure(false);
+  const userDetails = useAtomValue(userDetailsAtom);
+  const setOrganization = useSetAtom(organizationAtom);
+  const [emails, setEmails] = useState<string[]>([]);
+  const [emailInputValue, setEmailInputValue] = useState<string>('');
+  const [selectedRole, setSelectedRole] = useState<OrganizationRoleEnum>(OrganizationRoleEnum.ORG_MEMBER);
+  const [isPending, setIsPending] = useState<boolean>(false);
+  const rolesData = [
```
trixobird, line 35: this should be conditional based on the current user role. Admins should be able to invite admins and lesser, while Moderators should be able to invited moderators and lesser

trixobird (approved):
with a small comment

PR #334, opened 2024-08-10
Absolute values filtering insights

elevchyt:
Not sure if weekly filtering has been implemented on the BE, I've commented out on the FE.
If it works @trixobird you can uncomment it :)

Also, I didn't have insights (again) to test out how the dates look for different intervals, but filtering seems to work successfully.


apps/web/src/app/(authenticated)/insights/components/order-filters.tsx
```diff
@@ -59,6 +60,15 @@ export default function OrderFilters(): React.ReactNode {
     return isParamInSearchParams(searchParams, fetchPreviewsKey, 'true');
   };
 
+  const getIntervalValue = (): string => {
+    if (isParamInSearchParams(searchParams, intervalKey, InsightsInterval.month)) {
```
trixobird, line 64: this does not work for some reason. It always asks for week interval. You can verify it if you log in `insights-operations.ts` `args.filter.interval`

trixobird (commented):
now with the absolute option we may have insights with a single datapoint. This will create some issue here: https://github.com/adsviewer/turboviewer/blob/37d5f26d20d961abb7df6dda22c4e2876a790983/apps/web/src/app/(authenticated)/insights/components/insight-card.tsx#L75

apps/web/src/middleware.ts
```diff
@@ -112,7 +112,7 @@ export async function middleware(request: NextRequest): Promise<NextResponse> {
 
   // (Insights only) If page is loaded without any query params, set the following initial group filters
   if (request.nextUrl.pathname === '/insights' && !request.nextUrl.search) {
-    const newURL = `/insights?${groupedByKey}=${InsightsColumnsGroupBy.adId}&${groupedByKey}=${InsightsColumnsGroupBy.device}&${groupedByKey}=${InsightsColumnsGroupBy.publisher}&${groupedByKey}=${InsightsColumnsGroupBy.position}&${fetchPreviewsKey}=true`;
+    const newURL = `/insights?${groupedByKey}=${InsightsColumnsGroupBy.adId}&${groupedByKey}=${InsightsColumnsGroupBy.device}&${groupedByKey}=${InsightsColumnsGroupBy.publisher}&${groupedByKey}=${InsightsColumnsGroupBy.position}&${fetchPreviewsKey}=true&${intervalKey}=${InsightsInterval.month}`;
```
trixobird, line 115:
```suggestion
    const newURL = `/insights?${groupedByKey}=${InsightsColumnsGroupBy.adId}&${groupedByKey}=${InsightsColumnsGroupBy.device}&${groupedByKey}=${InsightsColumnsGroupBy.publisher}&${groupedByKey}=${InsightsColumnsGroupBy.position}&${fetchPreviewsKey}=true&${intervalKey}=${InsightsInterval.week}`;
```
let's keep the default as it is (week)

trixobird (approved):

PR #342, opened 2024-08-14
Update ad accounts

elevchyt:
Org admins can update the ad accounts that will be used for each integration.

Members & Operators may only view those ad accounts.


trixobird (commented):
image
I am the admin of this account, but this is a root organization. I should not be able to edit any of the accounts.

apps/server/src/schema/organization/organization-operations.ts
trixobird, line 1: image
I am a moderator/user to this account. I should be able to see one account (Happy_harbor)

trixobird, line 1: image
I am admin in this account (another account). No idea why it has different behavior on this account :D

trixobird (commented):
couple of functional issues. Also let's rebase so that we have the invite flow on this flow as well

trixobird:
Verified that in root organization has the Ad icon and behaves as it should. I created a sub account, to verify that the ad account will also be visible and if I am an admin I should be able to view the available and update.

The icon was not there in the sub organization. Maybe I am missing something though

trixobird:
I went to a sub org and I invited a user. After that the user table is not populated. The BE calls seems to work. More info: https://www.loom.com/share/5a3893a08ac946ca91504ba8b7e5bc71?sid=a2aff553-843f-470a-a42f-26681423d0ad

trixobird:
In an sub-org I correctly update the ad account (great :) ); I think it would be better if the card is refreshed with the active ads. More info here: https://www.loom.com/share/082136e582a54aeb8b94cedfe8fd2a40?sid=1662f701-903f-49f7-8cc0-008a33ba6125

elevchyt:
> In an sub-org I correctly update the ad account (great :) ); I think it would be better if the card is refreshed with the active ads. More info here: https://www.loom.com/share/082136e582a54aeb8b94cedfe8fd2a40?sid=1662f701-903f-49f7-8cc0-008a33ba6125

Went with reload for now. At some point we should definitely migrate the integrations to use Jotai like insights so that we can perform state-dependent updates on elements.

PR #358, opened 2024-08-15
integration token expiration notification & luxon install

elevchyt:
Shows a warning notification (that doesn't autoclose) to users regarding the expiration of their integrations' token.
Also, added a second button to each connected integration in the integrations page that allows both Revoke and Reconnect actions.


apps/web/src/util/integration-utils.ts
```diff
@@ -0,0 +1,14 @@
+import { DateTime } from 'luxon';
+import { type Integration } from '@/graphql/generated/schema-server';
+
+const TOKEN_WARNING_DAYS_THRESHOLD = 10;
+
+export const isTokenCloseToExpiration = (accessTokenExpiresAt: Integration['accessTokenExpiresAt']): boolean => {
+  if (accessTokenExpiresAt) {
```
trixobird, line 7: I just noticed that `accessTokenExpiresAt` is not deterministic on whether an integration is expiring. If it has a refresh token, then that's what we care about.

I am thinking to introduce a new state in the integration besides Expired, Connected etc called Expiring, what do you think?

apps/web/src/app/(authenticated)/integrations/components/integration-card.tsx
```diff
@@ -82,19 +83,39 @@ export default function IntegrationCard(props: IntegrationProps): ReactNode {
...
-            disabled={!isOrgAdmin(userDetails.allRoles) || !userDetails.currentOrganization?.isRoot}
+        <Flex direction="column" gap="xs" mt="xs">
+          <Tooltip
+            label={tGeneric('accessOrgAdminRoot')}
+            disabled={isOrgAdmin(userDetails.allRoles) && userDetails.currentOrganization?.isRoot}
           >
-            {t('revoke')}
-          </Button>
-        </Tooltip>
+            <Link href={props.authUrl ?? ''} passHref>
```
trixobird, line 91: is there a chance to make it to stand out in case the integration is Expiring?

trixobird (commented):
looks good, but unfortunately our initial idea about the expire date is not deterministic. I have added a comment with a suggestion

apps/web/src/util/integration-utils.ts
elevchyt: Yes! That would be much better, too!

trixobird: all right, I'll push something on your branch (probably tomorrow)

trixobird: done, we may not need luxor anymore, I added a helper function in date-utils.ts that you FE has also access to for future reference.

apps/web/src/app/(authenticated)/integrations/components/integration-card.tsx
elevchyt: Added an exclamation mark on the Reconnect button if it's status is Expiring.
Might add a little wiggle periodic animation once I get into animation stuff.

apps/web/package.json
```diff
@@ -58,6 +58,7 @@
     "jotai": "^2.9.1",
     "js-cookie": "^3.0.5",
     "lodash": "^4.17.21",
+    "luxon": "^3.5.0",
```
trixobird, line 61: do we still needs this (and the @types)?

apps/web/src/app/(authenticated)/integrations/components/integration-card.tsx
trixobird: hmmm it seems that it disappears right now instead of showing the exclamation mark
image

trixobird (commented):
couple of comments. Also the build is failing, let me know if you need any help fixing it.

trixobird:
I opened a similar [issue](https://github.com/adsviewer/turboviewer/issues/376), you may want to tackle as part of this PR, jfyi

apps/web/package.json
elevchyt: Yeah, we'll definitely use luxon at some point.

trixobird (approved):

PR #398, opened 2024-08-21
Absolute values filtering insights

elevchyt:


apps/web/src/app/(authenticated)/insights/components/order-filters.tsx
```diff
@@ -195,6 +195,21 @@ export default function OrderFilters(): React.ReactNode {
             maw={150}
             disabled={isPending}
           />
+          <Select
```
trixobird, line 198: this is already in this file (this does not have the `scrollAreaProps={{ type: 'always', offsetScrollbars: 'y' }}`

trixobird (commented):
it seems that we don't need this branch, the changes are already in main

PR #406, opened 2024-08-23
404-emulate-organization-admins

trixobird:
## Description

Clone the web app and then made modifications in order to have a single page emulate organizations

How to test: Make sure that you have an amin users (you can add the appropriate entry in the user_roles table). Then navigate to http://localhost:3002/. You will be redirected back to http://localhost:3000 to signin if you are not already and then you will see the emulate organization page. Just click on the drop down menu select and org and click on the button. You should be redirected to the web app as that user


elevchyt (approved):
Let's talk more about this on a call :)

PR #409, opened 2024-08-24
376 introduce errored integration

elevchyt:
Shows error badge on integration and notifies the user about the errored integration similar to the expiring token.


trixobird (approved):
This looks ok, It would be nice if the notification was clickable and the user would be directed to integrations.
If that's not an easy task, let's leave it for now.

elevchyt:
> This looks ok, It would be nice if the notification was clickable and the user would be directed to integrations. If that's not an easy task, let's leave it for now.

Not possible with natively using Mantine unfortunately. Doable but have to create a custom component for the notifications to do that.

PR #420, opened 2024-08-25
save tokens to web app when emulate

trixobird:
## Description

* web app sign-out is a route now
* in backoffice if user is not admin sign them out


apps/web/src/app/api/auth/sign-out/route.ts
elevchyt, line 1: Nice :)

elevchyt (approved):

PR #433, opened 2024-08-27
172 locale and language split

elevchyt:
Formatting will now be based on the user's browser locale. Was simpler than I thought :)

_The native NEXT_LOCALE entity hasn't been changed._


apps/web/src/app/user-locale-setter.tsx
```diff
@@ -0,0 +1,13 @@
+'use client';
+
+import { useEffect } from 'react';
+import { setCookie } from '@/app/actions';
+
+export default function UserLocaleSetter(): React.ReactNode {
+  useEffect(() => {
+    void setCookie('USER_LOCALE', navigator.language);
```
trixobird, line 8:
```suggestion
    Cookies.set('USER_LOCALE', navigator.language);
```
I am getting an error in console with the action
`Uncaught (in promise) Error: Only plain objects, and a few built-ins, can be passed to Client Components from Server Components. Classes or null prototypes are not supported.`

The same is happening in `color-scheme-picker.tsx`

apps/web/src/app/layout.tsx
```diff
@@ -8,22 +8,28 @@ import { NextIntlClientProvider } from 'next-intl';
...
 
 export default async function RootLayout({ children }: { children: React.ReactNode }): Promise<React.ReactNode> {
+  const headersList = headers();
+  const acceptLanguageHeaderValue = headersList.get('accept-language');
+  const preferredLocale = acceptLanguageHeaderValue ? acceptLanguageHeaderValue.split(',')[0] : 'en';
   const locale = await getLocale();
   const messages = await getMessages();
 
   return (
     <html lang={locale}>
```
trixobird, line 28:
```suggestion
    <html lang={preferredLocale}>
```
Is this correct? If yes we can remove the getLocale function

elevchyt: Done

trixobird (approved):

PR #450, opened 2024-08-28
date picker on insights

elevchyt:
Added a date picker on insights with **dateFrom** and **dateTo** params that is limited to having its latest date to today.

### DIFF
--- a/apps/web/src/app/(authenticated)/insights/components/order-filters.tsx
+++ b/apps/web/src/app/(authenticated)/insights/components/order-filters.tsx
@@ -5,6 +5,8 @@ import { usePathname, useRouter, useSearchParams } from 'next/navigation';
 import { useTranslations } from 'next-intl';
 import { type ChangeEvent, useTransition } from 'react';
 import { useSetAtom } from 'jotai/index';
+import { DatePickerInput, type DateValue, type DatesRangeValue } from '@mantine/dates';
+import { IconCalendarMonth } from '@tabler/icons-react';
 import {
   OrderDirection,
   addOrReplaceURLParams,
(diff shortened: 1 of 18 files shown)


apps/web/src/app/(authenticated)/insights/components/order-filters.tsx
```diff
@@ -109,6 +121,29 @@ export default function OrderFilters(): React.ReactNode {
...
+    const newParams = new URLSearchParams(searchParams.toString());
+    const dateFrom = dates[0]?.getTime();
+    const dateTo = dates[1]?.getTime();
+    if (dateFrom && dateTo) {
+      resetInsights();
+      newParams.set(dateFromKey, String(dateFrom));
+      newParams.set(dateToKey, String(dateTo));
+      const newURL = `${pathname}?${newParams.toString()}`;
+      startTransition(() => {
+        router.replace(newURL);
```
trixobird, line 134:
```suggestion
        window.history.replaceState(null, '', newURL)
```
Here are the docs: https://nextjs.org/docs/app/building-your-application/routing/linking-and-navigating#using-the-native-history-api you can replace the url without reloading the page.

Which means that if we have the insights call in a useEffect that is dependent only on searchparams, we should be re-rendering only the changed component and the insights area.

apps/web/src/app/(authenticated)/insights/components/order-filters.tsx
```diff
@@ -179,22 +214,37 @@ export default function OrderFilters(): React.ReactNode {
...
+                ]}
+                value={getIntervalValue()}
+                onChange={handleIntervalChange}
+                allowDeselect={false}
+                comboboxProps={{ transitionProps: { transition: 'fade-down', duration: 200 } }}
+                scrollAreaProps={{ type: 'always', offsetScrollbars: 'y' }}
+                maw={150}
+                disabled={isPending}
+              />
+              <DatePickerInput
```
trixobird, line 237: When I change locale to en-US the week does not start on Sunday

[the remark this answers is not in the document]
elevchyt: Fixed!

apps/web/src/app/(authenticated)/insights/components/order-filters.tsx
elevchyt: HUGE

elevchyt: Actually, I tried it on the page size dropdown and seems to be rerendering other stuff too :/
Also, doesn't seem to re-fetch data and shows "no results".

trixobird: Yes I know there are some issues, but I am assuming it is this neighbouring issue that triggers the re-render. I don't know what would be causing the no re-fetch though. We probably need to test this on a clean project

trixobird (approved):

PR #456, opened 2024-08-28
device on insight card

elevchyt:
Devices are now shown on insight cards.


elevchyt:
> could we have icon instead of text (we wouldn't need to translate as well) for example this for mobile? https://tabler.io/icons/icon/device-mobile

Done

trixobird (approved):

PR #474, opened 2024-08-30
444-sub-organizations-should-not-be-able-to-create-new-sub-sub-organizations

trixobird:


apps/web/src/components/shells/main-shell/main-shell.tsx
```diff
@@ -66,7 +66,7 @@ export function MainAppShell({ children }: { children: React.ReactNode }): React
           </Flex>
           <Flex align="center" justify="flex-end" ml="auto" gap="sm">
             <OrganizationSelect />
-            <CreateOrganizationButton />
+            {userDetails.currentOrganization?.isRoot ? <CreateOrganizationButton /> : null}
```
trixobird, line 69: I was thinking that we disable the button not hide it, and add a tool tip expaling that "Sub-sub-organizations are not allowed"

trixobird (commented):
Now if I don't have an org (delete the current org) I cannot create one

elevchyt:
> Now if I don't have an org (delete the current org) I cannot create one

Apologies, didn't think of that case. Fixed.

PR #480, opened 2024-08-31
fix landing page build issue

trixobird:
## Description

there is one thing that I cannot fix. In the send message the checkbox itself is not clickable (the label is). If I put the input inside the label works, not sure if there is a better solution


elevchyt:
If it works it's fine :)

elevchyt (approved):

PR #496, opened 2024-09-04
feat: allow specific max users per Tier

### LINKED ISSUE #495: Allow specific max users per Tier
As in landing page suggest we need to have **max** user per Tier.

Let's create maybe a map in packages/utils so we can share these rules with the FE. There are two places that a user is created, you can find by searching `prisma.user.create`.

Update: backed is ready. Front end should also make sure that if the org is already full the invite button should be disabled, maybe with a tooltip expalining why.

You probably would want to use [tier-mappings.ts](https://github.com/adsviewer/turboviewer/blob/62ecfefeac5ef5b13518991e6ad44ebfb0cd6b1c/packages/mappings/src/tier-mapping.ts#L13-L12).

aaryansinha16:
## Description

- We need to add max limit in the tiers according to the landing page

## Demo/Screenshots (If Applicable)

image

Closes #495


trixobird (commented):
good job you are learning :) 

some comments... If any of them are not clear please ping me in slack .

packages/utils/src/tier-rules.ts
```diff
@@ -0,0 +1,17 @@
+export enum Tier {
```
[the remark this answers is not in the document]
aaryansinha16: Since we are not able to import the database mappings in the utils file, should I omit this file completely?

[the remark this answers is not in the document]
trixobird: yes, we shouldn't add anything on utils

apps/server/src/contexts/user/user-invite.ts
```diff
@@ -94,6 +95,23 @@ export const createInvitedUser = async (
...
+  const organization = await prisma.organization.findUniqueOrThrow({
+    where: { id: organizationId },
+  });
+
+  const userOrganizationCount = await prisma.userOrganization.count({
+    where: { organizationId },
+  });
+
+  const currentTier = organization.tier;
+  const currentUserCount = userOrganizationCount;
```
trixobird, line 107: let's inline this, applicable in more places in this PR

apps/server/src/schema/user/invite-operations.ts
```diff
@@ -142,7 +143,7 @@ builder.mutationFields((t) => ({
 
       const frontEndInvitedUserUrl = new URL(`${env.PUBLIC_URL}/sign-up`);
 
-      const [usersMap, organization] = await Promise.all([
+      const [usersMap] = await Promise.all([
```
trixobird, line 146: why we removing the organization here, you need it down the line.

apps/server/src/schema/user/invite-operations.ts
```diff
@@ -158,6 +159,24 @@ builder.mutationFields((t) => ({
         args.emails.map(async (email) => {
           const inviteHash = generateConfirmInvitedUserToken();
           const dbUser = usersMap.get(email);
+
+          const organization = await prisma.organization.findUniqueOrThrow({
```
trixobird, line 163: you already had this on line 145

apps/server/src/schema/user/invite-operations.ts
```diff
@@ -214,23 +233,35 @@ builder.mutationFields((t) => ({
...
-                lastName: newUser.lastName,
-                email,
-                organizationName: organization.name,
-                expirationInDays: invitationExpirationInDays,
-                actionUrl: frontEndInvitedUserUrl.toString(),
-              }),
-              setConfirmInvitedUserRedis(inviteHash, newUser.id, ctx.organizationId),
-            ]);
+            const newUser: AError | User = await createInvitedUser(email, emailType, args.role, ctx.organizationId);
+            if (!(newUser instanceof AError)) {
```
trixobird, line 237: let's invert the if to avoid identation.
```
if (isAError(newUser)) return...

File: apps/server/src/schema/user/invite-operations.ts [hunk 6]
```diff
@@ -214,23 +233,35 @@ builder.mutationFields((t) => ({
...
+                  lastName: newUser.lastName,
+                  email,
+                  organizationName: organization.name,
+                  expirationInDays: invitationExpirationInDays,
+                  actionUrl: frontEndInvitedUserUrl.toString(),
+                }),
+                setConfirmInvitedUserRedis(inviteHash, newUser.id, ctx.organizationId),
+              ]);
+            } else {
+              return new AError('Error creating invited user');
```
trixobird, line 250:
```suggestion
              return { message: newUser.message, email };
```
This is what the errors expect

apps/server/src/schema/user/invite-operations.ts
```diff
@@ -214,23 +233,35 @@ builder.mutationFields((t) => ({
...
         }),
       );
-      const flattenErrors = errors.flatMap((e) => e ?? []);
+
+      function isEmailMessageObject(
+        error: AError | { email: string; message: string },
+      ): error is { email: string; message: string } {
+        return !(error instanceof AError);
+      }
+      const flattenErrors = errors.flatMap((e) => e ?? []).filter(isEmailMessageObject);
```
trixobird, line 261:
```suggestion
     const flattenErrors = errors.flatMap((e) => e ?? []);
```
you don't need these any more, errors return correct type

apps/server/src/contexts/user/user-invite.ts
aaryansinha16:
```typescript
const currentUserCount = await prisma.userOrganization.count({
  where: { organizationId },
});

const currentTier = (
  await prisma.organization.findUniqueOrThrow({
    where: { id: organizationId },
  })
).tier;
``` 

By inline you are referring to something like this ^?

trixobird: inline means that instead of using the variable `currentUserCount` in your case, to use whatever is on the right part of the equation `userOrganizationCount` in our case

apps/server/src/schema/user/invite-operations.ts
aaryansinha16: image
Error: 
`Argument of type '(AError | { email: string; message: string; })[]' is not assignable to parameter of type '{ email: string; message: string; }[]'.
  Type 'AError | { email: string; message: string; }' is not assignable to type '{ email: string; message: string; }'.
    Property 'email' is missing in type 'AError' but required in type '{ email: string; message: string; }`
 
 Had to typeguard due to the above error.

PR #497, opened 2024-09-04
446 sub organization include existing organization users

elevchyt:
This was a tough one...
Users may now add users from the current organization to a new organization they're creating.


elevchyt:
From a UX perspective: Maybe we should not show the current user that's creating the organization & maybe also filter out members whose status is still `INVITED`? @trixobird

elevchyt:
Done!

apps/web/src/app/(authenticated)/organization/actions.ts
```diff
@@ -62,28 +62,40 @@ async function deleteOrganization(
...
   try {
     const res = await createOrganization(values);
 
     if (!res.success) {
       logger.error(res.error);
-      return false;
+      return {
+        success: false,
+        message: res.error,
+      };
```
trixobird, line 77:
```suggestion
      return res;
```
I think it is the same signature with UrqlError. Also the logger.error is not needed, we log the error inside the urql. Applicable in one more place in the same file.

apps/web/src/components/create-organization/create-organization-button.tsx
```diff
@@ -9,32 +9,64 @@ import { useTranslations } from 'next-intl';
...
+    for (const user of newUsers) {
+      if (user.userId)
+        newUsersFiltered = [
+          ...newUsersFiltered,
+          {
+            userId: user.userId,
+            role: user.role,
+          },
+        ];
+    }
```
trixobird, line 47:
```suggestion
    const newUsersFiltered: UserRolesInput[] = newUsers
  .filter(user => user.userId)
  .map(user => ({
    userId: user.userId,
    role: user.role,
  }));
```

trixobird (commented):
Functionality is there, I am putting some Ux thought in this loom: https://www.loom.com/share/5a3e2d9f551647b9a26a2431134864ff

apps/web/src/app/(authenticated)/organization/actions.ts
elevchyt: Removed the logs.
UrqlErrorResult seems to be a bit different, but in my attempt to use it, it expects <TError> argument which I couldn't exactly import into the actions.ts file.

elevchyt: Fixed.

elevchyt:
Added the logged in user as the default user & disabled the Add Users button (with a hint) on orgs that have only 1 member (the logged in user).

apps/web/src/app/(authenticated)/organization/actions.ts
trixobird: still see the object instead of `return res` :P

apps/web/src/app/(authenticated)/organization/actions.ts
```diff
@@ -62,55 +62,56 @@ async function deleteOrganization(
...
 
     const switchRes = await switchOrganization({ organizationId: res.data.createOrganization.id });
 
     if (!switchRes.success) {
-      logger.error(switchRes.error);
-      return false;
+      return {
+        success: false,
+        message: switchRes.error,
+      };
```
trixobird, line 85:
```suggestion
      return switchRes;
```

apps/web/src/components/create-organization/add-users-modal.tsx
```diff
@@ -0,0 +1,248 @@
...
+};
+
+export default function AddUsersModal(props: PropsType): ReactNode {
+  const t = useTranslations('organization');
+  const tGeneric = useTranslations('generic');
+  const tProfile = useTranslations('profile');
+  const [opened, { open, close }] = useDisclosure(false);
+  const [organization, setOrganization] = useAtom(organizationAtom);
+  const userDetails = useAtomValue(userDetailsAtom);
+  // const usersData: UserRolesInput[] = [];
```
trixobird, line 42:
```suggestion
```

trixobird (approved):
couple of small comments

PR #513, opened 2024-09-10
499 limit integration and added switchTier endpoint

aaryansinha16:
## Description

We  limit the integrations to 1 when the tier is Launch.
Created an endpoint called `switchTiers` which would handle  upgrade/downgrades of the tiers


apps/server/src/schema/organization/organization-operations.ts
```diff
@@ -359,6 +359,110 @@ builder.mutationFields((t) => ({
...
+        }
+      }
+
+      const updatedOrganization = await prisma.organization.update({
+        ...query,
+        where: { id: organizationId },
+        data: { tier },
+      });
+
+      return updatedOrganization;
```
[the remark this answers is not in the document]
aaryansinha16: Should I create a separate function to disconnect the integrations, is there any other possible places where we can use that in future?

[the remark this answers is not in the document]
trixobird: separate function seems always a good idea :)

packages/mappings/src/tier-mapping.ts
```diff
@@ -1,7 +1,7 @@
 import { type Tier } from '@repo/database';
 
-export const maxUsersPerTier: Record<Tier, { maxUsers: number }> = {
-  Launch: { maxUsers: 1 },
+export const maxUsersPerTier: Record<Tier, { maxUsers: number; maxIntegrations?: number }> = {
+  Launch: { maxUsers: 1, maxIntegrations: 1 },
   Build: { maxUsers: 4 },
   Grow: { maxUsers: 10 },
   Scale: { maxUsers: 100 },
```
trixobird, line 7:
```suggestion
  Build: { maxUsers: 4, maxIntegrations: Infinity },
  Grow: { maxUsers: 10, maxIntegrations: Infinity },
  Scale: { maxUsers: 100, maxIntegrations: Infinity },
```
This way we have the business on this file, and we don't need to do `const maxIntegrations = maxUsersPerTier[tierStatus].maxIntegrations ?? Infinity;`

apps/server/src/schema/organization/organization-operations.ts
```diff
@@ -359,6 +359,22 @@ builder.mutationFields((t) => ({
...
+    args: {
+      organizationId: t.arg.string({ required: true }),
+      tier: t.arg({ type: Tier, required: true }),
+    },
+    resolve: async (query, _root, args, _ctx, _info) => {
+      const { organizationId, tier: newTier } = args;
+
+      const updatedOrganization = switchTierHelper(organizationId, newTier, query);
+
+      return updatedOrganization;
```
trixobird, line 375:
```suggestion
      return switchTierHelper(organizationId, newTier, query);
```
let's try to return immediately when possible.

packages/organization/src/organization.ts
```diff
@@ -0,0 +1,144 @@
...
+
+export const getRootOrganizationId = async (organizationId: string): Promise<string> => {
+  const organization = await prisma.organization.findUniqueOrThrow({ where: { id: organizationId } });
+  if (organization.parentId) {
+    return await getRootOrganizationId(organization.parentId);
+  }
+  return organization.id;
+};
+
+export const getTier = async (organizationId?: string | null): Promise<Tier> => {
```
trixobird, line 16: I'd prefer to have organizationId only as string. In the only case that is null (on the sign up) let's make the check there. If no organizationId we return Launch or else we call the getTier

packages/channel/src/integration-helper.ts
```diff
@@ -88,6 +90,20 @@ const completeIntegration = async (
...
+  const tierStatus = await getTier(organizationId);
+  const maxIntegrations = maxUsersPerTier[tierStatus].maxIntegrations ?? Infinity;
+
+  const integrations = await prisma.integration.findMany({
+    where: {
+      organizationId,
+    },
+  });
+
+  const currentIntegrations = integrations.reduce((acc, el) => (el.status === 'CONNECTED' ? acc + 1 : acc), 1);
```
trixobird, line 102:
```suggestion
    const currentIntegrations = await prisma.integration.count({
    where: {
      organizationId,
      status: IntegrationStatus.CONNECTED,
    },
  });
```

packages/organization/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "@repo/organization",
+  "version": "1.0.0",
+  "private": true,
+  "description": "Shared Organization Utilities. Use only types from package/database",
```
trixobird, line 5:
```suggestion
  "description": "Shared Organization Utilities.
```

packages/organization/src/organization.ts
```diff
@@ -0,0 +1,144 @@
...
+  if (!organizationId) {
+    return Tier.Launch;
+  }
+
+  const rootOrganizationId = await getRootOrganizationId(organizationId);
+  const organization = await prisma.organization.findUniqueOrThrow({ where: { id: rootOrganizationId } });
```
trixobird, line 22:
```suggestion
    const organization = await prisma.organization.findUniqueOrThrow({
    select: { tier: true }, where: { id: rootOrganizationId },
  });
```
This is jfyi to learn if you haven't already that you can do this on prisma. You may keep the code as is.

packages/organization/src/organization.ts
```diff
@@ -0,0 +1,144 @@
...
+      const adminUsers = orgUsers
+        .filter((userOrg) => userOrg.role === 'ORG_ADMIN')
+        .sort((a, b) => new Date(a.user.createdAt).getTime() - new Date(b.user.createdAt).getTime());
+
+      const regularUsers = orgUsers
+        .filter((userOrg) => userOrg.role !== 'ORG_ADMIN')
+        .sort((a, b) => new Date(a.user.createdAt).getTime() - new Date(b.user.createdAt).getTime());
+
+      const usersToKeep = [
+        ...adminUsers.slice(0, maxUsersInTier),
```
trixobird, line 87: we don't need to keep all administrators, just one. So I think what we need to do is get all the users order by date and slice them. If there is an amin among them, we are cool. If not, we need to remove the newest user and add the oldest admin.

packages/organization/src/organization.ts
```diff
@@ -0,0 +1,144 @@
...
+    ...organizationWithUsers,
+  });
+
+  const currentTier = organization.tier;
+
+  if (currentTier === newTier) {
+    return organization;
+  }
+
+  const disconnectIntegration = async (orgId: string): Promise<void> => {
```
trixobird, line 44: let's put all the internal functions at the beginning of the `swtichTierHelper` function. Right now the flow is interupted by the internal function.

packages/organization/src/organization.ts
```diff
@@ -0,0 +1,144 @@
...
+      ...query,
+      where: { id: organizationId },
+      data: { tier: newTier },
+    }),
+
+    prisma.organization.updateMany({
+      where: { parentId: organizationId },
+      data: { tier: newTier },
+    }),
+  ]);
```
trixobird, line 130:
```suggestion
    await prisma.organization.update({
    ...query,
    where: { id: organizationId },
    data: { 
      tier: newTier,
      children: {
        updateMany: {
          where: { parentId: organizationId },
          data: { tier: newTier },
        },
      }
    },
  });
```
I know that I said that this was the only way; I played around a bit and I think this should work (we need to verify though)

packages/organization/src/organization.ts
```diff
@@ -0,0 +1,144 @@
...
+      await prisma.userOrganization.deleteMany({
+        where: {
+          userId: { in: usersToRemove.map((userOrg) => userOrg.userId) },
+          organizationId: orgId,
+        },
+      });
+    }
+  };
+
+  await removeUserFromOrganization(organizationId, organization.users, newTier);
```
trixobird, line 102: I think this won't make the cut. Let's say that we switch from Growth to Build. The org has 5 users in the root organization, user A(admin), B/C/D,F(members) and 2 users in subOrg userF(admin) and userE. Let's assume that the letters indicate when they are created so A is the oldest, E is the most recent.

After the switch the whole organization should have 4 users right? They should be the oldest users, with one of them being an org admin. So A,B,C,D.

The code as I undertand it would keep all 6 of the users.

Now in this case the subOrg would be without any user. We need to mindful of it and assign the oldest admin of root org as admin in the subOrg

packages/organization/src/organization.ts
trixobird, line 44: I don't understand what the code is doing. It should dsconnect all integrations, but the oldest no? I think it disconnects only the oldest.

packages/organization/src/organization.ts
```diff
@@ -0,0 +1,144 @@
...
+  };
+
+  const removeUserFromOrganization = async (
+    orgId: string,
+    orgUsers: OrganizationWithUsers[`users`],
+    tier: Tier,
+  ): Promise<void> => {
+    const maxUsersInTier = maxUsersPerTier[tier].maxUsers;
+
+    if (tier === Tier.Launch || tier === Tier.Build || tier === Tier.Grow) {
```
trixobird, line 74: I think we can revert the if and remove identation.

packages/organization/src/organization.ts
```diff
@@ -0,0 +1,144 @@
...
+      if (tier === Tier.Launch) {
```
trixobird, line 75: you are hard coding on the organization.ts business code that is on Tier.  The whole if should be something like `await disconnectIntegration(maxUsersPerTier[tier].maxIntegrations, orgId);`

apps/server/src/schema/integrations/integration-operations.ts
```diff
@@ -47,17 +48,29 @@ builder.queryFields((t) => ({
           organizationId: rootOrgId,
         },
       });
+      const tierStatus = await getTier(ctx.organizationId);
+      const maxIntegrations = maxUsersPerTier[tierStatus].maxIntegrations ?? Infinity;
```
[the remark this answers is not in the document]
aaryansinha16: What do you think about `tierLimits` or `tierConstraints` ?

trixobird:
> tierConstraints :)
> 
Kind regards,
Giorgos Trichopoulos

apps/server/src/schema/organization/organization-operations.ts
```diff
@@ -359,6 +359,110 @@ builder.mutationFields((t) => ({
...
+            },
+          },
+        },
+      });
+
+      const currentTier = organization.tier;
+      const maxUsersInNewTier = maxUsersPerTier[tier].maxUsers;
+
+      if (currentTier !== tier) {
+        if (tier === 'Launch') {
```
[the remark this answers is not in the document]
trixobird: This issues still persists in line 161 of this file.

packages/organization/src/organization.ts
```diff
@@ -0,0 +1,159 @@
...
+      where: {
+        userId: { in: usersToRemove.map((userOrg) => userOrg.userId) },
+        organizationId: { in: organizationIds },
+      },
+    });
+
+    for (const subOrg of subOrganizations) {
+      const subOrgAdmin = subOrg.users.find((user) => user.role === 'ORG_ADMIN');
+      if (!subOrgAdmin) {
+        const rootOrgAdmin = usersToKeep.find((user) => user.role === 'ORG_ADMIN');
```
trixobird, line 97: let assign the oldest not admin, not a random one.

packages/organization/src/organization.ts
```diff
@@ -0,0 +1,159 @@
...
+        }
+      }
+    }
+  };
+
+  const organization = await prisma.organization.findUniqueOrThrow({
+    ...query,
+    where: { id: organizationId },
+    ...organizationWithUsers,
+  });
```
trixobird, line 112:
```suggestion
  const organizations = await prisma.organization.findMany({
    where: { OR: [{id: organizationId}, {parentId: organizationId}] },
    ...organizationWithUsers,
  });
```
I was thiking that maybe we don't need to differentiate between orgs and suborgs. All of the will have the same Tier. And in all orgs we need to do the same. Remove the extra users.

avoid using the ...query unless you are returning the object at the last of the call. There maybe uneeded joins in there.

trixobird (commented):
we are closing...

packages/organization/src/organization.ts
```diff
@@ -0,0 +1,159 @@
...
+    if (tier !== Tier.Launch && tier !== Tier.Build && tier !== Tier.Grow) {
+      return;
+    }
+
+    await disconnectIntegration(maxUsersInTier, orgId);
+
+    const allUsers = [...orgUsers];
+    subOrganizations.forEach((subOrg) => allUsers.push(...subOrg.users));
+
+    const sortedUsers = allUsers.sort(
```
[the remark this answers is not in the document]
aaryansinha16: Thank you for the great suggestion! Seeing the practical use of the Set data structure here is awesome.

packages/organization/src/organization.ts
aaryansinha16: A bit of confusion on the `not admin`,, can you please clarify?
We need to get the oldest admin for rootOrgAdmin and subOrgAdmin right? (in rootOrgAdmin we are already getting that, since the `usersToKeep` comes from `sortedUsers` & I'll update the part for subOrgAdmin)

packages/organization/src/organization.ts
```diff
@@ -0,0 +1,148 @@
...
+      const integrationsToDisconnect = integrations.slice(maxIntegrations);
+
+      integrationsToDisconnect.forEach((integration) => {
+        fireAndForget.add(() => revokeIntegration(integration.externalId, integration.type));
+      });
+    }
+  };
+  const removeUserFromOrganization = async (
+    orgId: string,
+    allUsers: OrganizationWithUsers[`users`],
```
trixobird, line 48:
```suggestion
    allUsers: OrganizationWithUsers['users'],
```

packages/organization/src/organization.ts
```diff
@@ -0,0 +1,148 @@
...
+    tier: Tier,
+    subOrganizations: OrganizationWithUsers[],
```
trixobird, line 50:
```suggestion
    organizations: OrganizationWithUsers[],
```

packages/organization/src/organization.ts
```diff
@@ -0,0 +1,148 @@
...
+      }
+    }
+  };
+
+  const organizations = await prisma.organization.findMany({
+    where: { OR: [{ id: organizationId }, { parentId: organizationId }] },
+    ...organizationWithUsers,
+  });
+
+  const currentTier = organizations[0].tier;
```
trixobird, line 108: this will blow if organizations are empty

packages/organization/src/organization.ts
```diff
@@ -0,0 +1,148 @@
...
+  if (currentTier === newTier) {
+    return organizations[0];
+  }
+
+  const subOrganizations = organizations.filter((org) => org.parentId === organizationId);
+
+  const allUsers = organizations.flatMap((orgItem) => orgItem.users);
+  const maxIntegrations = tierConstraints[newTier].maxIntegrations;
+
+  await removeUserFromOrganization(organizationId, allUsers, newTier, subOrganizations);
```
trixobird, line 119:
```suggestion
  await removeUserFromOrganization(allUsers, newTier, organizations);
```
Why are you passing subOrganization here? Shouldn't you remove the users from the root organization as well?

packages/organization/src/organization.ts
```diff
@@ -0,0 +1,148 @@
...
+  const currentTier = organizations[0].tier;
+
+  if (currentTier === newTier) {
+    return organizations[0];
+  }
+
+  const subOrganizations = organizations.filter((org) => org.parentId === organizationId);
+
+  const allUsers = organizations.flatMap((orgItem) => orgItem.users);
+  const maxIntegrations = tierConstraints[newTier].maxIntegrations;
```
trixobird, line 117: let's move this under line 119 since it is related to line 120. Let's not confuse flows.

packages/organization/src/organization.ts
```diff
@@ -0,0 +1,148 @@
...
+
+        if (rootOrgAdmin) {
+          await prisma.userOrganization.update({
+            where: { userId_organizationId: { userId: rootOrgAdmin.userId, organizationId: subOrg.id } },
+            data: { role: OrganizationRoleEnum.ORG_ADMIN },
+          });
+        }
+      }
+    }
+  };
```
trixobird, line 101:
```suggestion
    for (const org of subOrganizations) {
      const keptUsers = usersToKeep.filter((user) => user.organizationId === org.id);
      if (!keptUsers.some((u) => u.role === OrganizationRoleEnum.ORG_ADMIN) && oldestAdmin) {
        await prisma.userOrganization.update({
          where: { userId_organizationId: { userId: oldestAdmin.userId, organizationId: org.id } },
          data: { role: OrganizationRoleEnum.ORG_ADMIN },
        });
      }
    }
```
do you see any bug here? You will need to move oldestOrganization outside the if.

packages/organization/src/organization.ts
trixobird: my mistake sorry

packages/organization/src/organization.ts
aaryansinha16: Can there be a case where we don't have any organization at all? We do have one by default too right?

trixobird: you are right :)

trixobird (approved):
we did it :)

PR #526, opened 2024-09-16
512 Added max-recency boundries

aaryansinha16:
### Added the data limit for maximum recency allowed by each tier

### DIFF
--- a/packages/channel-utils/src/date-utils.ts
+++ b/packages/channel-utils/src/date-utils.ts
@@ -1,27 +1,36 @@
 import { addInterval, getDayPriorTillTomorrow, getLastXMonths, getTomorrowStartOfDay } from '@repo/utils';
-import { prisma } from '@repo/database';
+import { prisma, Tier } from '@repo/database';
 import { tierConstraints } from '@repo/mappings';
 
 export const timeRanges = async (initial: boolean, adAccountId: string): Promise<{ since: Date; until: Date }[]> => {
-  if (initial) {
-    const range = getLastXMonths();
-    return splitTimeRange(range.since, range.until);
-  }
   const latestInsight = await prisma.insight.findFirst({
     select: { date: true, adAccount: { select: { organizations: { select: { tier: true } } } } },
     where: { adAccountId },
     orderBy: { date: 'desc' },
   });
-
-  const organizationTier = latestInsight?.adAccount?.organizations[0]?.tier;
-
-  const furthestDate = addInterval(new Date(), 'day', -tierConstraints[organizationTier || 'Launch'].maxRecency);
+  
+  const organizations = latestInsight?.adAccount?.organizations ?? [];
+  
+  const organizationWithHighestTier = organizations.length > 0
+    ? organizations.sort((a, b) => tierConstraints[b.tier].order - tierConstraints[a.tier].order)[0].tier
+    : undefined;
+  
+  if (initial) {
+    const range = getLastXMonths();
+    
+    if (!organizationWithHighestTier) {
+      return splitTimeRange(tierConstraints[Tier.Launch].maxRecency, range.since, range.until);
+    }
+  
+    return splitTimeRange(tierConstraints[organizationWithHighestTier].maxRecency, range.since, range.until);
+  }
+  const furthestDate = addInterval(new Date(), 'day', -tierConstraints[organizationWithHighestTier ?? Tier.Launch].maxRecency);
   const range = latestInsight ? getDayPriorTillTomorrow(latestInsight.date) : getDayPriorTillTomorrow(furthestDate);
-  return splitTimeRange(range.since, range.until, tierConstraints[organizationTier || 'Launch'].maxRecency);
+  return splitTimeRange(tierConstraints[organizationWithHighestTier ?? Tier.Launch].maxRecency, range.since, range.until);
 };
 
-export const splitTimeRange = (since: Date, until: Date = new Date(), maxTimePeriodDays?: number): { since: Date; until: Date }[] => {
-  const maxTimePeriod = 1000 * 60 * 60 * 24 * (maxTimePeriodDays || 7); // min recency in days
+export const splitTimeRange = (maxTimePeriodDays: number, since: Date, until: Date = new Date()): { since: Date; until: Date }[] => {
+  const maxTimePeriod = 1000 * 60 * 60 * 24 * maxTimePeriodDays;
   if (since.getTime() > new Date().getTime()) return [];
 
   const periods: { since: Date; until: Date }[] = [];
@@ -29,9 +38,9 @@ export const splitTimeRange = (since: Date, until: Date = new Date(), maxTimePer
   const diff = until.getTime() - since.getTime();
   if (diff < maxTimePeriod) return [{ since, until }];
 
-  const newDateUntil = addInterval(since, 'day', (maxTimePeriodDays || 7) - 1);
+  const newDateUntil = addInterval(since, 'day', (maxTimePeriodDays) - 1);
   periods.push({ since, until: newDateUntil });
   const newDateSince = getTomorrowStartOfDay(newDateUntil);
-  periods.push(...splitTimeRange(newDateSince, until));
+  periods.push(...splitTimeRange(maxTimePeriod, newDateSince, until));
   return periods;
 };
--- a/packages/channel-utils/src/date-utils.ts
+++ b/packages/channel-utils/src/date-utils.ts
@@ -1,4 +1,4 @@
-import { addInterval, getDayPriorTillTomorrow, getLastXMonths, getTomorrowStartOfDay } from '@repo/utils';
+import { addInterval, getDayPriorTillTomorrow, getLastXMonths } from '@repo/utils';
 import { prisma, Tier } from '@repo/database';
 import { tierConstraints } from '@repo/mappings';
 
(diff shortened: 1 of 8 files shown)


packages/channel-utils/src/date-utils.ts
```diff
@@ -1,32 +1,35 @@
...
   }
   const latestInsight = await prisma.insight.findFirst({
-    select: { date: true },
+    select: { date: true, adAccount: { select: { organizations: { select: { tier: true } } } } },
     where: { adAccountId },
     orderBy: { date: 'desc' },
   });
-  const furthestDate = addInterval(new Date(), 'day', -maxTimePeriodDays);
+
+  const organizationTier = latestInsight?.adAccount?.organizations[0]?.tier;
```
trixobird, line 16: we talked about it :) Why are getting the first organization? What if the user belong to different organization with different Tiers?

packages/mappings/src/tier-mapping.ts
```diff
@@ -1,12 +1,12 @@
...
-export const maxUsersPerTier: Record<Tier, { maxUsers: number }> = {
-  Launch: { maxUsers: 1 },
-  Build: { maxUsers: 4 },
-  Grow: { maxUsers: 10 },
-  Scale: { maxUsers: 100 },
+export const tierConstraints: Record<Tier, { maxUsers: number, maxRecency: number }> = {
+  Launch: { maxUsers: 1, maxRecency: 7 },
+  Build: { maxUsers: 4, maxRecency: 90 },
+  Grow: { maxUsers: 10, maxRecency: 365 },
+  Scale: { maxUsers: 100, maxRecency: Infinity },
```
trixobird, line 7:
```suggestion
  Scale: { maxUsers: 100, maxRecency: 365 },
```
Let's put 365, infinite is a dangerous setting

packages/channel-utils/src/date-utils.ts
```diff
@@ -1,32 +1,35 @@
...
   const range = latestInsight ? getDayPriorTillTomorrow(latestInsight.date) : getDayPriorTillTomorrow(furthestDate);
-  return splitTimeRange(range.since, range.until);
+  return splitTimeRange(range.since, range.until, tierConstraints[organizationTier || 'Launch'].maxRecency);
 };
 
-const maxTimePeriodDays = 90; // 90 days
-const maxTimePeriod = 1000 * 60 * 60 * 24 * maxTimePeriodDays;
-export const splitTimeRange = (since: Date, until: Date = new Date()): { since: Date; until: Date }[] => {
+export const splitTimeRange = (since: Date, until: Date = new Date(), maxTimePeriodDays?: number): { since: Date; until: Date }[] => {
+  const maxTimePeriod = 1000 * 60 * 60 * 24 * (maxTimePeriodDays || 7); // min recency in days
```
trixobird, line 24: let's not hard code 7 we have this info on the `tierConstraints`

trixobird (commented):
this should work, let's fix the comments. You also have some eslint errors. You probably need to configure vscode to show them. Mine for example are 
image

packages/channel-utils/src/date-utils.ts
```diff
@@ -1,32 +1,35 @@
...
+  const furthestDate = addInterval(new Date(), 'day', -tierConstraints[organizationTier || 'Launch'].maxRecency);
   const range = latestInsight ? getDayPriorTillTomorrow(latestInsight.date) : getDayPriorTillTomorrow(furthestDate);
-  return splitTimeRange(range.since, range.until);
+  return splitTimeRange(range.since, range.until, tierConstraints[organizationTier || 'Launch'].maxRecency);
 };
 
-const maxTimePeriodDays = 90; // 90 days
-const maxTimePeriod = 1000 * 60 * 60 * 24 * maxTimePeriodDays;
-export const splitTimeRange = (since: Date, until: Date = new Date()): { since: Date; until: Date }[] => {
+export const splitTimeRange = (since: Date, until: Date = new Date(), maxTimePeriodDays?: number): { since: Date; until: Date }[] => {
```
aaryansinha16, line 23:
```suggestion
export const splitTimeRange = (since: Date, until: Date = new Date(), maxTimePeriodDays: number = 7): { since: Date; until: Date }[] => {
```
How about using this, because if we use the tierContraints value rather then hardcoding 7 below in the function, we'll need to pass another optional parameter.... maybe `tier` or `minRecency`.

trixobird: I think it is essential for this functiona to have a tier as parameter yes.

packages/channel-utils/src/date-utils.ts
```diff
@@ -1,24 +1,49 @@
...
-import { prisma } from '@repo/database';
+import { prisma, Tier } from '@repo/database';
+import { tierConstraints } from '@repo/mappings';
 
 export const timeRanges = async (initial: boolean, adAccountId: string): Promise<{ since: Date; until: Date }[]> => {
-  if (initial) {
-    const range = getLastXMonths();
-    return splitTimeRange(range.since, range.until);
-  }
   const latestInsight = await prisma.insight.findFirst({
```
trixobird, line 6: if we don't have an insight this will return null, therefore no organizations. I think we should do a separate call for the organizations.

By the way the _AdAccountToOrganization table is populated when you connect to an integration. Since we never connect, but we hard code the access_token, we need to populate this table manually :(

packages/channel-utils/src/date-utils.ts
```diff
@@ -1,24 +1,49 @@
...
+    }
+
+    return splitTimeRange(tierConstraints[organizationWithHighestTier].maxRecency, range.since, range.until);
+  }
+  const furthestDate = addInterval(
+    new Date(),
+    'day',
+    -tierConstraints[organizationWithHighestTier ?? Tier.Launch].maxRecency,
+  );
   const range = latestInsight ? getDayPriorTillTomorrow(latestInsight.date) : getDayPriorTillTomorrow(furthestDate);
```
trixobird, line 33: let's say that we have an integration with the last insight 9 months back, (maybe the user revoked the integration or  it has enabled another integration due to tier contraints. 

Let's say that he paying for build Tier (3 months of insights). They way that this code reads, I think it will get insights for 9 months not 3.

trixobird (commented):
let's create some tests for the `timeRanges`. In order not o mess with db calls let's create another function that start after the db call.

Let me know if you want to talk about it.

packages/channel-utils/src/date-utils.ts
aaryansinha16: _AdAccountToOrganization table is automatically generated by prisma since adAccount and organization have many to many relationship. 
By manually populating, do you mean something like this:

```typescript
const updatedAdAccount = await prisma.adAccount.update({
    where: { id: adAccountId },
    data: {
      organizations: {
        connect: { id: organizationId },
      },
    },
  });

Comment (trixobird) [reply packages/channel-utils/src/date-utils.ts:6]: hmmm scratch that comment. The code that connects the 2 [runs ](https://github.com/adsviewer/turboviewer/blob/62fa50c0ca11a02f175dbc5e164f4b56c5f7c3cd/packages/channel-utils/src/insights-utils.ts#L7-L31)on every refresh, so we should have the connection (but I think we don't). Can you make some tests and see if this code is working?

File: packages/channel-utils/src/date-utils.ts [hunk 7]
```diff
@@ -1,24 +1,53 @@
...
+    if (!organizationWithHighestTier) {
+      return splitTimeRange(tierConstraints[Tier.Launch].maxRecency, range.since, range.until);
+    }
+
+    return splitTimeRange(maxRecency, range.since, range.until);
+  }
+  const furthestDate = addInterval(new Date(), 'day', -maxRecency);
+  const insightDateWithinLimit = latestInsight && latestInsight.date > furthestDate ? latestInsight.date : furthestDate;
+  const range = getDayPriorTillTomorrow(insightDateWithinLimit);
+  return splitTimeRange(maxRecency, range.since, range.until);
```
trixobird, line 42:
```suggestion
  return testTimeRange(initial, latestInsight);
```
I am a bit confused. You have created a `testTimeRange` (probably a name that we need to change) function that you test, but the product code does not use it.

packages/channel-utils/test/date-util.test.ts
```diff
@@ -1,44 +1,76 @@
 import { describe, it } from 'node:test';
 import * as assert from 'node:assert';
-import { splitTimeRange } from '../src/date-utils';
+import { Tier } from '@repo/database';
+import { splitTimeRange, testTimeRange } from '../src/date-utils';
```
trixobird, line 4:
```suggestion
import { splitTimeRange, testTimeRange } from '../src';
```

packages/channel-utils/test/date-util.test.ts
```diff
@@ -57,4 +94,31 @@ void describe('splitTimeRange tests', () => {
...
   });
+
+  void it('returns split time range when passed with false initial', () => {
+    const dummyInsight = {
+      date: new Date('2024-06-25T18:30:00.000Z'),
+      adAccount: { organizations: [{ tier: Tier.Build }] },
+    };
+
+    const ranges = testTimeRange(false, dummyInsight);
+    assert.strictEqual(ranges.length, 2);
```
trixobird, line 105: it seems that these tests will fail as time passes. [This](https://nodejs.org/api/test.html#dates) should help you.

packages/channel-utils/test/date-util.test.ts
```diff
@@ -57,4 +94,31 @@ void describe('splitTimeRange tests', () => {
     assert.strictEqual(ranges[2].since.toISOString(), '2021-06-30T00:00:00.000Z');
     assert.strictEqual(ranges[2].until.toISOString(), '2021-09-25T00:00:00.000Z');
   });
+
+  void it('returns split time range when passed with false initial', () => {
```
trixobird, line 98: Could we have a test where the last insight was 3 months back and the tier is Launch?

packages/channel-utils/test/date-util.test.ts
```diff
@@ -57,4 +94,50 @@ void describe('splitTimeRange tests', () => {
...
+
+    const ranges = timeRangeHelper(true, dummyInsight);
+    assert.strictEqual(ranges.length, 2);
+    assert.strictEqual(ranges[0].since.toISOString(), '2024-01-01T00:00:00.000Z');
+    assert.strictEqual(ranges[0].until.toISOString(), '2024-03-30T00:00:00.000Z');
+    assert.strictEqual(ranges[1].since.toISOString(), '2024-03-31T00:00:00.000Z');
+    assert.strictEqual(ranges[1].until.toISOString(), '2024-06-26T18:30:00.000Z');
+  });
+
+  void it('returns correct time range when last insight was 3 months ago and tier is Launch', (context) => {
```
trixobird, line 128: you have a Lauch tier that according to our [site](https://adsviewer.io/#pricing) it should have only the last 7 days. But we are returning timeRanges that are from the beginning of the year. This does not seems correct.

packages/channel-utils/test/date-util.test.ts
```diff
@@ -57,4 +94,50 @@ void describe('splitTimeRange tests', () => {
...
+      date: new Date('2024-03-25T18:30:00.000Z'),
+      adAccount: { organizations: [{ tier: Tier.Launch }] },
+    };
+
+    const ranges = timeRangeHelper(true, dummyInsight);
+
+    assert.strictEqual(ranges.length, 26);
+    assert.strictEqual(ranges[0].since.toISOString(), '2024-01-01T00:00:00.000Z');
+    assert.strictEqual(ranges[0].until.toISOString(), '2024-01-07T00:00:00.000Z');
+    assert.strictEqual(ranges[1].since.toISOString(), '2024-01-08T00:00:00.000Z');
```
trixobird, line 141: it is find not to check all 26 ranges, but please check the first and the last at least.

packages/channel-utils/test/date-util.test.ts
trixobird, line 128: I want another test like this with initial false

packages/channel-utils/test/date-util.test.ts
```diff
@@ -57,4 +94,63 @@ void describe('splitTimeRange tests', () => {
...
+      adAccount: { organizations: [{ tier: Tier.Build }] },
+    };
+
+    const ranges = timeRangeHelper(true, dummyInsight);
+    assert.strictEqual(ranges.length, 1);
+    assert.strictEqual(ranges[0].since.toISOString(), '2024-03-28T18:30:00.000Z');
+    assert.strictEqual(ranges[0].until.toISOString(), '2024-06-25T18:30:00.000Z');
+  });
+
+  void it('returns correct time range when last insight was 3 months ago and tier is Launch', (context) => {
```
trixobird, line 126: these last two test should produce the same result. I would expect one range from midnight 18 till 26 midnight.

trixobird (approved):

PR #533, opened 2024-09-21
improved org data fetching via jotai & added max integrations logic p…

elevchyt:
Users will now have the integrations' "Connect" button disabled if they have reached (or exceeded) the maximum allowed amount of integrations for their pricing plan's tier.

Also had to make some changes for loading the initial organization data.


apps/web/src/components/user-button/user-button.tsx
```diff
@@ -51,11 +54,21 @@ export default function UserButton(): ReactNode {
         checkIntegrationTokensForExpiration(res.currentOrganization?.integrations as Integration[] | null);
         setUserDetails(res);
         setIsDataLoaded(true);
+
+        void getOrganization()
```
trixobird, line 58: why do we need this here?

apps/web/src/app/(authenticated)/integrations/components/integrations-grid.tsx
```diff
@@ -86,6 +90,13 @@ export default function IntegrationsGrid(props: IntegrationProps): ReactNode {
     }
   }, [props.metadata]);
 
+  const allowConnectionBasedOnTier = (): boolean => {
```
trixobird, line 93: Some comments assuming we are at Lauch tier:
- if user is above the allowed integration revoke should work.
- if user has 3 integrations but all of them are not connected all btns should work
- if user has 1 integration connected (but maybe more integrations in non connected state) the btns of the 1 integration should work.

apps/web/src/components/user-button/user-button.tsx
elevchyt: fixed by adding tier to the current org data of the me query

apps/web/src/app/(authenticated)/integrations/components/integrations-grid.tsx
elevchyt: I kinda rushed this one, sorry...
Tested all of the above cases and it now works as intended.

Also fixed a sneaky layout bug that allowed users to click over the disabled buttons and bypass the "Connect" button's disabled state.

apps/web/src/components/user-button/user-button.tsx
```diff
@@ -3,21 +3,23 @@
...
 import { IntegrationStatus, type Integration } from '@/graphql/generated/schema-server';
+import { organizationAtom } from '@/app/atoms/organization-atoms';
 import classes from './user-button.module.scss';
 
 export default function UserButton(): ReactNode {
   const tGeneric = useTranslations('generic');
   const tIntegrations = useTranslations('integrations');
   const router = useRouter();
   const [userDetails, setUserDetails] = useAtom(userDetailsAtom);
+  const setOrganization = useSetAtom(organizationAtom);
```
trixobird, line 22: do we need this now? (and the other changes in this file)

apps/web/src/app/(authenticated)/integrations/components/integrations-grid.tsx
```diff
@@ -86,6 +90,17 @@ export default function IntegrationsGrid(props: IntegrationProps): ReactNode {
     }
   }, [props.metadata]);
 
+  const allowNewConnectionBasedOnTier = (): boolean => {
+    if (userDetails.currentOrganization?.tier) {
+      if (
+        userDetails.currentOrganization.integrations.length >=
```
trixobird, line 96:
```suggestion
        userDetails.currentOrganization.integrations.filter((i) => i.status === IntegrationStatus.Connected).length >=
```

apps/web/src/components/user-button/user-button.tsx
elevchyt: fixed

trixobird (approved):

PR #540, opened 2024-09-22
show ad previews position change

elevchyt:
UI improvement for the "Show ad previews" switch.

### DIFF
--- a/apps/web/src/app/(authenticated)/insights/components/order-filters.tsx
+++ b/apps/web/src/app/(authenticated)/insights/components/order-filters.tsx
@@ -43,16 +43,13 @@ export default function OrderFilters(): React.ReactNode {
   };
 
   const getPageSizeValue = (): string => {
-    if (isParamInSearchParams(searchParams, urlKeys.pageSize, searchParams.get(urlKeys.pageSize) ?? '12')) {
+    if (isParamInSearchParams(searchParams, urlKeys.pageSize, searchParams.get(urlKeys.pageSize) ?? '12'))
       return searchParams.get(urlKeys.pageSize) ?? '12';
-    }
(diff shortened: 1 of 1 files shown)


apps/web/src/app/(authenticated)/insights/components/order-filters.tsx
```diff
@@ -163,24 +156,36 @@ export default function OrderFilters(): React.ReactNode {
   const handleAdPreviewChange = (e: ChangeEvent<HTMLInputElement>): void => {
     resetInsights();
     let newURL: string;
-    if (e.target.checked) {
-      newURL = addOrReplaceURLParams(pathname, searchParams, urlKeys.fetchPreviews, 'true');
-    } else {
-      newURL = addOrReplaceURLParams(pathname, searchParams, urlKeys.fetchPreviews);
-    }
+    if (e.target.checked) newURL = addOrReplaceURLParams(pathname, searchParams, urlKeys.fetchPreviews, 'true');
+    else newURL = addOrReplaceURLParams(pathname, searchParams, urlKeys.fetchPreviews);
```
trixobird, line 160:
```suggestion
    const newURL = e.target.checked
      ? addOrReplaceURLParams(pathname, searchParams, urlKeys.fetchPreviews, 'true')
      : addOrReplaceURLParams(pathname, searchParams, urlKeys.fetchPreviews);
```
also in general we want to keep the brackets on if/else

you may even do:
```
    const newURL = addOrReplaceURLParams(
      pathname,
      searchParams,
      urlKeys.fetchPreviews,
      e.target.checked ? 'true' : undefined,
    );
```

trixobird (approved):

Issue #511, opened 2024-09-08
Time constraints Bugs


dennislobster:
Couple of strange combinations which don't return results:

[] Quarterly - never works
![Screenshot 2024-09-08 at 16 39 14](https://github.com/user-attachments/assets/9008dd37-efb0-4484-98e3-919617f7a1ca)

[] Daily - does not work with date filter also applied
![Screenshot 2024-09-08 at 16 37 40](https://github.com/user-attachments/assets/3791924f-a802-4bee-b85a-00ce227dafe0)

Generally it seems that a granularity overrides the time range, ie. selecting 1 day of data and changing monthly, weekly, quarterly gives different values.

PR #544, opened 2024-09-23
511 time contraints bug fixes

aaryansinha16:
This PR contains fixes of 2 bugs related to time constraints filters.

Closes: #511


trixobird (commented):
let's write some tests that would fail with the old code, but now are passing. We want test that would cover as much of the code as possible.

This is one of the, if not 'the' core business of adsviewer, we need to make sure that is correct.

Also make sure that you rebase, I made some changes recently

Issue #543, opened 2024-09-23
User feedback functionality


elevchyt:
A feedback button should be implemented so that users can provide useful feedback and/or bug reports.
There should be a relation in the DB between the user and the feedback entry, a **feedback type** field with the values: "Bug Report", "Feature Suggestion" and "Other", and also another field for the feedback's **message body**.

Ideally, the users should receive an email informing them that their feedback was submitted successfully.

trixobird:
should we close this @aaryansinha16 , @elevchyt ?

PR #545, opened 2024-09-23
543 added sendFeedback endpoint

aaryansinha16:
## Description
This PR adds the `sendFeedback` endpoint which the frontend can use, for users to send feedbacks. 
There are 3 types of feedbacks : `Bug_Report`, `Feature_Suggestion`, `Other`

Makes progress in #543


apps/server/src/schema/feedback/feedback-operation.ts
```diff
@@ -0,0 +1,55 @@
...
+import { FireAndForget } from '@repo/utils';
+import { GraphQLError } from 'graphql';
+import { builder } from '../builder';
+import { sendFeedbackReceivedEmail } from '../../email';
+import { FeedbackDto, FeedbackTypeEnum } from './feedback-types';
+
+const fireAndForget = new FireAndForget();
+
+builder.mutationFields((t) => ({
+  sendFeedback: t.withAuth({ authenticated: true }).field({
```
trixobird, line 11:
```suggestion
  sendFeedback: t.withAuth({ authenticated: true }).prismaField({
```
If you return the result of a prisma operation, let's return a prisma field

apps/server/src/schema/feedback/feedback-operation.ts
```diff
@@ -0,0 +1,55 @@
...
+    type: FeedbackDto,
+    args: {
+      message: t.arg.string({ required: true }),
+      type: t.arg({
+        type: FeedbackTypeEnum,
+        required: true,
+      }),
+    },
+    validate: (args) => Boolean(args.message && args.type),
+    resolve: async (_root, args, ctx, _info) => {
```
trixobird, line 22:
```suggestion
    resolve: async (query, _root, args, ctx, _info) => {
```
since we are returning a prismaField, pothos add query as the first parameter of resolve

apps/server/src/schema/feedback/feedback-operation.ts
```diff
@@ -0,0 +1,55 @@
...
+      const feedback = await prisma.feedback.create({
+        data: {
+          userId: ctx.currentUserId,
+          message,
+          type,
+          createdAt: new Date(),
+        },
+      });
+
+      fireAndForget.add(() => sendFeedbackReceivedEmail(user.email, user.firstName, user.lastName));
```
trixobird, line 44: you can move this up, and return directly from the prisma.feedback.create

apps/server/src/schema/feedback/feedback-types.ts
```diff
@@ -0,0 +1,18 @@
+import { builder } from '../builder';
+
+export const FeedbackTypeEnum = builder.enumType('FeedbackType', {
+  values: {
+    BUG_REPORT: { description: 'Report a bug' },
+    FEATURE_SUGGESTION: { description: 'Suggest a feature' },
+    OTHER: { description: 'Other feedback' },
+  },
+});
```
trixobird, line 9:
```suggestion
export const FeedbackTypeEnum = builder.enumType(FeedbackType, {
  name: 'FeedbackTypeEnum',
});
```
We don't need cat is cat descriptions https://www.reddit.com/r/ProgrammerHumor/comments/8w54mx/code_comments_be_like/

apps/server/src/schema/feedback/feedback-types.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const FeedbackDto = builder.simpleObject('Feedback', {
```
trixobird, line 11:
```suggestion
export const FeedbackDto = builder.prismaObject('Feedback', {
```
We always prioritze prisma object when possible. Look around to see its syntax

apps/web/src/util/schemas/user-schemas.ts
```diff
@@ -0,0 +1,6 @@
+import { z } from 'zod';
+
+export const SendFeedbackSchema = z.object({
+  type: z.string(),
+  message: z.string().min(24, { message: 'Message must be at least 24 characters' }),
```
trixobird, line 5: @elevchyt let's put this in utils so both web and server have access. Like we have done with passwordSchema.
@aaryansinha16 let's use this validation in the message. Look how we used the passwordSchema, should be the same.

packages/database/prisma/schema.prisma
```diff
@@ -79,6 +80,21 @@ model User {
   @@map("users")
 }
 
+enum FeedbackType {
```
trixobird, line 83:
```suggestion
enum FeedbackTypeEnum {
```

packages/database/prisma/schema.prisma
```diff
@@ -79,6 +80,21 @@ model User {
...
+  BUG_REPORT
+  FEATURE_SUGGESTION
+  OTHER
+}
+
+model Feedback {
+  id        String       @id @default(cuid())
+  user      User         @relation(fields: [userId], references: [id])
+  userId    String
```
trixobird, line 92:
```suggestion
  userId    String @map("user_id")
```
we use the snake case in db

packages/database/prisma/schema.prisma
```diff
@@ -79,6 +80,21 @@ model User {
...
+  type      FeedbackType
+  message   String
+  createdAt DateTime     @default(now()) @map("created_at")
+}
+
```
trixobird, line 97:
```suggestion
@@map("feedbacks")

```
we use snake_case in db, and plural.

apps/web/src/components/buttons/feedback-button.tsx
trixobird, line 1: This is for @dennis-adsviewer. Is this sufficient? I was thinking something more intrucive, like a floating button, butom right (that maybe has a close btn that sends it on present location, with a cool animation even :P sorry Lefteri, I was carried away :P )
image

elevchyt: My initial thought was also to place it on the bottom right, but we should reserve that for live support (or chatbot support if we ever have that xd)

apps/web/src/util/schemas/user-schemas.ts
trixobird: I was referring only the the message. the type should be `type: z.nativeEnum(FeedbackType),` an enum that we don't have in utils

apps/web/src/components/buttons/feedback-button.tsx
dennis-adsviewer: I think this button is okay, but we should in the future build function to for example highlight it when a new user has not yet used it, etc.

elevchyt: I'm planning to turn it in to a multi option button once things start getting more cluttered and we'll have only one button which opens a dropdown with the send feedback functionality and other stuff that we might add.

packages/database/prisma/migrations/20240927093413_feedbacks/migration.sql
```diff
@@ -0,0 +1,31 @@
...
+    "user_id" TEXT NOT NULL,
+    "type" "FeedbackTypeEnum" NOT NULL,
+    "message" TEXT NOT NULL,
+    "created_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
+    CONSTRAINT "feedbacks_pkey" PRIMARY KEY ("id")
+);
+
+-- AddForeignKey
+ALTER TABLE "feedbacks" ADD CONSTRAINT "feedbacks_user_id_fkey" FOREIGN KEY ("user_id") REFERENCES "users"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
```
trixobird, line 31:
```suggestion
-- Rename userId
ALTER TABLE "Feedback" RENAME COLUMN "userId" TO "user_id";

-- Rename Table
ALTER TABLE "Feedback" RENAME TO "feedbacks";
```
don't trust blindly the sql generation of prisma, especially when you rename things.

packages/database/prisma/migrations/20240927093413_feedbacks/migration.sql
```diff
@@ -0,0 +1,31 @@
+/*
+  Warnings:
+
+  - You are about to drop the `Feedback` table. If the table is not empty, all the data it contains will be lost.
+
+*/
```
trixobird, line 6:
```suggestion
```

packages/utils/src/shared-schemas.ts
```diff
@@ -0,0 +1,8 @@
+import { z } from 'zod';
+
+export const SendFeedbackSchema = z.object({
```
trixobird, line 3:
```suggestion
export const sendFeedbackSchema = z.object({
```

packages/utils/src/shared-schemas.ts
```diff
@@ -0,0 +1,8 @@
...
+  type: z.string(),
+  message: z.string().min(24, { message: 'Message must be at least 24 characters' }),
+});
+
+export const MessageSchema = z.string().min(24, { message: 'Message must be at least 24 characters' });
```
trixobird, line 8:
```suggestion
export const messageSchema = z.string().min(24, { message: 'Message must be at least 24 characters' });
```

apps/server/src/schema/feedback/feedback-operation.ts
trixobird: `add query as the first parameter of resolve` @aaryansinha16

packages/utils/src/shared-schemas.ts
```diff
@@ -0,0 +1,8 @@
+import { z } from 'zod';
+
+export const SendFeedbackSchema = z.object({
+  type: z.string(),
+  message: z.string().min(24, { message: 'Message must be at least 24 characters' }),
```
trixobird, line 5:
```suggestion
  message: messageSchema,
```
Hmmm I think sometimes I don't manage to pass through to you @aaryansinha16 what is in my mind :) You are duplicating the `z.string().min(24, { message: 'Message must be at least 24 characters' })` which beats the purpose of my initial comment https://github.com/adsviewer/turboviewer/pull/545#discussion_r1777989104

aaryansinha16: I agree, that was a silly duplication :sweat_smile:

trixobird (commented):
missing [this](https://github.com/adsviewer/turboviewer/pull/545#discussion_r1779979232) comment , and then we are done :)

apps/server/src/schema/feedback/feedback-operation.ts
```diff
@@ -0,0 +1,36 @@
...
+      }),
+    },
+    resolve: async (query, _root, args, ctx, _info) => {
+      const { message, type } = args;
+
+      const user = await prisma.user.findUniqueOrThrow({ where: { id: ctx.currentUserId } });
+
+      fireAndForget.add(() => sendFeedbackReceivedEmail(user.email, user.firstName, user.lastName));
+
+      return await prisma.feedback.create({
```
trixobird, line 27:
```suggestion
      return await prisma.feedback.create({
              ...query,
```
You are right, I didn't notice that we haven't exposed everything in FeedbackDto. I added the relevant code.

you are not using the query variable from the resolve. That is code smell, we always need to use this to enable graphql joins. For example this:
```
mutation sendFeedback {
  sendFeedback(message: "da", type: BUG_REPORT) {
    user {
      id
    }
  }
}
```
won't be able to work without the `...query`

apps/server/src/schema/feedback/feedback-types.ts
```diff
@@ -0,0 +1,16 @@
...
+  name: 'FeedbackTypeEnum',
+});
+
+export const FeedbackDto = builder.prismaObject('Feedback', {
+  fields: (t) => ({
+    type: t.expose('type', {
+      type: FeedbackType,
+      nullable: false,
+    }),
+    message: t.exposeString('message', { nullable: false }),
```
trixobird, line 14:
```suggestion
    message: t.exposeString('message', { nullable: false }),
    createdAt: t.expose('createdAt', { type: 'Date', nullable: false }),
    userId: t.exposeID('userId', { nullable: false }),
    user: t.expose('user', { nullable: false, type: UserDto }),
```
In general we are exposing all the things from schema. Consumer is the one that chooses what are the things that will actuall fetch.

aaryansinha16: In the `user` field, let's use `t.relation` instead. 
`t.expose` won't work as it does not directly exists in the prisma model

trixobird (approved):

PR #558, opened 2024-10-01
525 summary view

aaryansinha16:
## Description
This PR lifts getInsights logic to a helper function in packages and is being re-used in the data-refresh.ts


packages/channel/src/insights-helper.ts
```diff
@@ -0,0 +1,95 @@
...
+    selections: {
+      kind: string;
+      name: { value: string };
+    }[];
+  };
+}
+
+export const getInsightsHelper = async (
+  filter?: FilterInsightsInputType, // TODO: should be required
+  organizationId?: string, // TODO: should be required
```
trixobird, line 29:
```suggestion
  filter: FilterInsightsInputType,
  organizationId: string,
```
let's make the mandatory from now.

packages/channel/src/data-refresh.ts
```diff
@@ -59,6 +60,8 @@ export const refreshData = async ({
       await refreshDataOf(integration, initial);
       deleteInsightsCache(integration.organizationId);
     }
+
+    await getInsightsHelper();
```
trixobird, line 64: you can get the organizatId here. If there are integrationIds, you need to find the distinct organizationIds of these integrations, and if there are no integrationIds, you need to loop through all the organizations.

lets use 
```
{
      interval: 'week', orderBy: 'spend_abs', page: 0, pageSize: 3
    }
    ``` for `FilterInsightsInputType` for now.

File: packages/channel/src/insights-helper.ts [hunk 3]
```diff
@@ -0,0 +1,95 @@
...
+  acceptedLocale?: string,
+): Promise<GroupedInsightsWithEdges> => {
+  // TODO :remove the condition and make the parameters required
+  if (!filter || !organizationId || !acceptedLocale) throw new AError('Required parameters not passed');
+
```
trixobird, line 34:
```suggestion
```

trixobird:
I rebased and did an extra commit, because not we are filtering on adIds and adAccountIds

trixobird:
by the way it seems that this PR is not grouping by integration (a.k.a channel) which was the main thing of this issue. Should we rename this issue and open another, or you will do it here?

packages/channel/src/data-refresh.ts
aaryansinha16: Lets say there are multiple organizationIds for all the different integrationIds, so should I loop through them and make request for every organizationId? Like this: 
```typescript
    for (const integration of integrations) {
      await getInsightsHelper({
        interval: 'week', orderBy: 'spend_abs', page: 0, pageSize: 3
      }, integration.organizationId);
    }
```

trixobird: that would execute the function multiple time for the same integration. Why don't you loop over the integrations, get their organizationIds as set (so uniqu) and then you loop through theses orgnizationIds?

aaryansinha16:
I'll do it here in the same issue itself.

apps/server/src/schema/integrations/insights-operations.ts
```diff
@@ -50,69 +46,14 @@ builder.queryFields((t) => ({
...
-      if (redisValue) return redisValue;
-
-      const groupBy: (InsightsColumnsGroupByType | 'currency')[] = [...(args.filter.groupBy ?? []), 'currency'];
-
-      const insightsRaw: Record<string, never>[] = await prisma.$queryRawUnsafe(
-        groupedInsights(args.filter, ctx.organizationId, ctx.acceptedLocale),
+      return getInsightsHelper(
+        args.filter,
+        ctx.organizationId,
+        Language[ctx.acceptedLocale.toUpperCase() as keyof typeof Language],
```
trixobird, line 55:
```suggestion
        ctx.acceptedLocale
```
We don't like the `as` because it is not gurantees types during run time. Here either we force the acceptedLocale in an enum from the very start. So [here](https://github.com/adsviewer/turboviewer/blob/81cdf8045ab50f80fc1f0d15d5d37867f30bc061/apps/server/src/language.ts#L18), we have a list of all possible languages or we keep it a string. For now let's keep it a string.

In general is better to type things as early as possible in the flow.

apps/server/src/schema/integrations/insights-operations.ts
```diff
@@ -247,7 +190,7 @@ const IFrameDTO = builder.objectRef<IFrameWithType>('IFrame').implement({
 });
 
 const GroupedInsightDto = builder.simpleObject(
-  'GroupedInsights',
+  'GroupedInsight',
```
trixobird, line 193: <3

packages/channel/src/data-refresh.ts
```diff
@@ -59,6 +60,19 @@ export const refreshData = async ({
...
     }
+
+    const organizationIds = new Set(integrations.map((integration) => integration.organizationId));
+    for (const organizationId of organizationIds) {
+      await getInsightsHelper(
+        {
+          interval: 'week',
+          orderBy: 'spend_abs',
+          page: 1,
+          pageSize: 3,
```
trixobird, line 71:
```suggestion
          pageSize: 3,
          groupBy: ['integrationType']
```
if we don't group my anything we will have the 3 top of the thousands insights of this org. If you group you get the top 3 out of the number of connected integrations.

trixobird (approved):
one minor comment :)

apps/server/src/schema/integrations/insights-operations.ts
aaryansinha16: Since this `GroupedInsightDto` represents a single object, I thought a singular term would be much better. Is that not the case?

PR #559, opened 2024-10-01
548 implement testing library

elevchyt:
Added Jest w/ configuration for Next.js, created a sample test and added /web test coverage prerequisite on Husky's pre-commit configuration.


apps/web/src/__tests__/page.test.tsx
trixobird, line 1: my only comment is that we are using a different naming convetion. We are useing ./tests/**/*.test.ts and you are using ./src/__tests__/*.test.ts

trixobird (approved):

apps/web/src/__tests__/page.test.tsx
elevchyt: That's because the sample test I made is for the root component `Home` which is at the root of the app.
Tests are going to follow the folder structure regularly.

This was my reference: https://github.com/vercel/next.js/tree/canary/examples/with-jest
As you can see, the __tests__ folder is inside the same folder as the app folder, we just happen to use the /src folder as a container (which was defined upon the initialization of the /web project).

PR #575, opened 2024-10-06
574 cookies consent google analytics setup

elevchyt:
Cookies consent & google analytics setup for the landing page.
You need to get a new .env for the `/landing-page` folder to make this work.


trixobird:
How come and we don't use nextjs official suggestion? https://nextjs.org/docs/messages/next-script-for-ga
How can I see the analytics? Which account is being used?
Is there a way to introduce different environmetns to the analytics, or does it understand by default due to different urls?

apps/landing-page/src/components/google-analytics.tsx
```diff
@@ -0,0 +1,35 @@
...
+        strategy="lazyOnload"
+        src={`https://www.googletagmanager.com/gtag/js?id=${String(process.env.NEXT_PUBLIC_MEASUREMENT_ID)}`}
+      />
+
+      <Script id="" strategy="lazyOnload">
+        {`
+              window.dataLayer = window.dataLayer || [];
+              function gtag(){dataLayer.push(arguments);}
+              gtag('js', new Date());
+              gtag('config', '${String(process.env.NEXT_PUBLIC_MEASUREMENT_ID)}', {
```
trixobird, line 28: let's use the `env.mjs` Since we don't care about local environment we can have a default value of `dont track`

elevchyt:
> How come and we don't use nextjs official suggestion? https://nextjs.org/docs/messages/next-script-for-ga How can I see the analytics? Which account is being used? Is there a way to introduce different environmetns to the analytics, or does it understand by default due to different urls?

Ohh pretty nice, I followed [this](https://medium.com/readytowork-org/google-analytics-in-next-js-a26cc2b28db5), but it is probably outdated.

dennislobster:
@trixobird @elevchyt we can also implement this for the web application, I created a new property for this in google analytics, we can use this tag / tag code for the app.adsviewer.io domain:


<script async src="https://www.googletagmanager.com/gtag/js?id=G-GZRN0Y7X1L"></script>
<script>
  window.dataLayer = window.dataLayer || [];
  function gtag(){dataLayer.push(arguments);}
  gtag('js', new Date());

  gtag('config', 'G-GZRN0Y7X1L');
</script>

apps/landing-page/src/components/cookies-consent.tsx
trixobird, line 1: do we need the same for the web?

trixobird (approved):

Issue #552, opened 2024-09-30
Feedback not sending to database


dennis-adsviewer:
When hitting submit on feedback it is not populating in feedbacks table.

Requests:
[] fix to post feedback to feedbacks table
[] disable sending email on submit
[] add animation (confetti?) to show that feedback was submitted and close feedback box
[] post in #feedback channel in slack (to create) when new feedback is added

trixobird:
so point 2 and 4 are for @aaryansinha16 and point 3 is for @elevchyt . I think these are independent things, so feel free to create seperate branches from fe/be

elevchyt:
3 should definitely be its own branch/issue since we're gonna need to implement [lottie](https://lottiefiles.com/) for those types of animations

PR #580, opened 2024-10-07
CurrentOrg in feedback & feedback posts on slack

aaryansinha16:
## Description
This PR addresses: 
- post in #public-feedback channel in slack (to create) when new feedback is added
- save the current organization of the user in the table.

Makes progress in #552


apps/server/src/schema/feedback/feedback-helper.ts
```diff
@@ -0,0 +1,22 @@
+import { type Feedback } from '@repo/database';
+import { AError } from '@repo/utils';
+
+export async function postFeedbackToSlack(feedback: Feedback): Promise<AError | undefined> {
+  const slackMessage = {
+    text: `New Feedback Received: \n*Type:* ${feedback.type} \n*Message:* ${feedback.message}`,
+  };
+
+  const slackWebhookUrl = process.env.SLACK_WEBHOOK_URL_PUBLIC_FEEDBACK ?? '';
```
trixobird, line 9: look at `apps/server/src/config.ts` we are using this pattern when we add env vars. Also we need to add this variable in `turbo.json`

apps/server/src/schema/feedback/feedback-helper.ts
```diff
@@ -0,0 +1,22 @@
+import { type Feedback } from '@repo/database';
+import { AError } from '@repo/utils';
+
+export async function postFeedbackToSlack(feedback: Feedback): Promise<AError | undefined> {
```
trixobird, line 4:
```suggestion
export const postFeedbackToSlack = async (feedback: Feedback): Promise<AError | undefined> => {
```

apps/server/src/schema/feedback/feedback-helper.ts
```diff
@@ -0,0 +1,22 @@
...
+  const slackMessage = {
+    text: `New Feedback Received: \n*Type:* ${feedback.type} \n*Message:* ${feedback.message}`,
```
trixobird, line 6:
```suggestion
    text: `New ${MODE !== Environment.Production ? 'Test ' : ''}Feedback Received: \n*Type:* ${feedback.type} \n*Message:* ${feedback.message}`,
```

apps/server/src/schema/feedback/feedback-helper.ts
```diff
@@ -0,0 +1,22 @@
...
+  try {
+    await fetch(slackWebhookUrl, {
+      method: 'POST',
+      body: JSON.stringify(slackMessage),
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    });
+  } catch (_err) {
+    return new AError('Something went wrong while sending message to slack');
```
trixobird, line 20:
```suggestion
    logger.error(e)
    return new AError('Something went wrong while sending message to slack');
```

apps/server/src/schema/feedback/feedback-helper.ts
```diff
@@ -0,0 +1,22 @@
+import { type Feedback } from '@repo/database';
+import { AError } from '@repo/utils';
+
+export async function postFeedbackToSlack(feedback: Feedback): Promise<AError | undefined> {
+  const slackMessage = {
+    text: `New Feedback Received: \n*Type:* ${feedback.type} \n*Message:* ${feedback.message}`,
+  };
+
+  const slackWebhookUrl = process.env.SLACK_WEBHOOK_URL_PUBLIC_FEEDBACK ?? '';
+
```
trixobird, line 10: let's return if there is no slackWebhookUrl

packages/database/prisma/schema.prisma
```diff
@@ -87,12 +88,14 @@ enum FeedbackTypeEnum {
...
-  userId    String           @map("user_id")
-  type      FeedbackTypeEnum
-  message   String
-  createdAt DateTime         @default(now()) @map("created_at")
+  id                    String           @id @default(cuid())
+  user                  User             @relation(fields: [userId], references: [id])
+  userId                String           @map("user_id")
+  type                  FeedbackTypeEnum
+  message               String
+  currentOrganization   Organization?    @relation(fields: [currentOrganizationId], references: [id])
```
trixobird, line 96: let's put the relation last with an empty line

apps/server/src/config.ts
```diff
@@ -9,6 +9,7 @@ const schema = z
     GOOGLE_APPLICATION_ID: z.string().min(1),
     GOOGLE_APPLICATION_SECRET: z.string().min(1),
     EMAILABLE_API_KEY: z.string().min(1),
+    SLACK_WEBHOOK_URL_PUBLIC_FEEDBACK: z.string().url(),
```
trixobird, line 12:
```suggestion
    SLACK_WEBHOOK_URL_PUBLIC_FEEDBACK: z.string().url().optional(),
```
or else if dev does not have this on their local .env the app won't start. Since it does not impact the core fucntionality of the app we can make it optionanl

apps/server/src/schema/feedback/feedback-helper.ts
```diff
@@ -0,0 +1,29 @@
...
+import { logger } from '@repo/logger';
+import { Environment, MODE } from '@repo/mode';
+import { AError } from '@repo/utils';
+
+export const postFeedbackToSlack = async (feedback: Feedback): Promise<AError | undefined> => {
+  const slackMessage = {
+    text: `New ${MODE !== Environment.Production ? 'Test ' : ''} Feedback Received: \n*Type:* ${feedback.type} \n*Message:* ${feedback.message}`,
+  };
+
+  const slackWebhookUrl = process.env.SLACK_WEBHOOK_URL_PUBLIC_FEEDBACK;
```
trixobird, line 11:
```suggestion
  const slackWebhookUrl = env.SLACK_WEBHOOK_URL_PUBLIC_FEEDBACK;
```
This is how you refer it now that you put it in config. Now it is typed as well

apps/server/src/schema/feedback/feedback-helper.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  if (!slackWebhookUrl) {
+    return new AError('No slack webhook URL provided');
+  }
```
trixobird, line 15:
```suggestion
  if (!slackWebhookUrl) return;
```
I think it is fine if we just return. This way devs without the URL will be able to test feedback without flooding the channel.

trixobird (approved):

PR #585, opened 2024-10-09
525 improved summary view

elevchyt:
This is a _relatively_ big one, since I had to tackle the way the navbar operates and make it handle sublinks as well.
Implemented the summary view, updated the way user details are fetched so that they're independent of the `user-button.tsx` since problems arose due to that.

Also updated some lottie graphics and added some juicy navlink/sublink & user profile button animations.

For some reason, using the day interval doesn't seem to return any insights at all, that's why you most likely won't be seeing the chart in summary.
(You can comment out the `interval` property from `chartParams` inside `chart-container.tsx` as a temp fix if needed for testing)


apps/web/src/app/(authenticated)/summary/components/top-ads-container.tsx
```diff
@@ -0,0 +1,188 @@
...
+            return res;
+          })
+          .catch((error: unknown) => {
+            logger.error(error);
+            return null;
+          })
+          .finally(() => {
+            setIsPending(false);
+          });
+        allRequests = [...allRequests, request];
```
trixobird, line 105:
```suggestion
        allRequests.push(request);
```
O(n) vs O(1)

apps/web/src/app/(authenticated)/summary/layout.tsx
```diff
@@ -0,0 +1,15 @@
+import React, { Suspense } from 'react';
+import { useTranslations } from 'next-intl';
+import { Title } from '@mantine/core';
+import LoaderCentered from '@/components/misc/loader-centered';
+
+export default function RootLayout({ children }: { children: React.ReactNode }): React.ReactNode {
```
trixobird, line 6:
```suggestion
export default function SummaryLayout({ children }: { children: React.ReactNode }): React.ReactNode {
```

trixobird (approved):

PR #610, opened 2024-10-14
fix linked in revoke functionality

trixobird:


trixobird:
@dennis-adsviewer this should be fixing the linked in revoke. (The others were fine when I tested them, let me know if you have different experiences)?

PR #628, opened 2024-10-18
551 google data ingress

aaryansinha16:


packages/channel-google/src/channel-google.ts
```diff
@@ -0,0 +1,775 @@
...
+    if (!integration.refreshToken) return new AError('Refresh token is required');
+
+    const ranges = await timeRanges(initial, dbAccount.id);
+
+    for (const range of ranges) {
+      try {
+        const refreshedIntegration = await Google.refreshedIntegration(integration);
+        if (isAError(refreshedIntegration)) return refreshedIntegration;
+
+        const url = `https://googleads.googleapis.com/${apiVersion}/customers/${dbAccount.externalId}/googleAds:search`;
```
trixobird, line 294: lets put that as baseUrl or similar

packages/channel-google/src/channel-google.ts
```diff
@@ -0,0 +1,775 @@
...
+            metrics.impressions,
+            metrics.clicks,
+            metrics.cost_micros,
+            
+            metrics.video_quartile_p25_rate,
+            metrics.video_quartile_p50_rate,
+            metrics.video_quartile_p75_rate,
+            metrics.video_quartile_p100_rate,
+            metrics.video_view_rate,
+            metrics.video_views,
```
trixobird, line 333: are you saving all these things? Do you need them? If not let's remove them.

packages/channel-google/src/channel-google.ts
```diff
@@ -0,0 +1,775 @@
...
+        segments.ad_format_type
+      FROM video
+      WHERE segments.date BETWEEN '${formatYYYMMDDDate(new Date(range.since))}' AND '${formatYYYMMDDDate(new Date(range.until))}'
```
trixobird, line 336:
```suggestion
      WHERE segments.date BETWEEN '${formatYYYMMDDDate(range.since)}' AND '${formatYYYMMDDDate(range.until)}'
```

packages/channel-google/src/channel-google.ts
```diff
@@ -0,0 +1,775 @@
...
+    `;
+
+        const youtubeData = await Google.handlePagination(
+          integration,
+          videoQuery,
+          VideoAdResponseSchema,
+          dbAccount.externalId,
```
trixobird, line 343: you don't need to change it now:

it is better to pass a processFn as well, that will save the youtubeData in every call. We are doing this in channel-meta. This way you don't create giant memory structures, that could in turn create other issues, and also you don't risk losing the data in a crash or similar.

packages/channel-google/src/channel-google.ts
```diff
@@ -0,0 +1,775 @@
...
+        );
+
+        if (!youtubeData) return;
+
+        if (isAError(youtubeData)) return;
```
trixobird, line 348:
```suggestion
if (!youtubeData || isAError(youtubeData)) return;
```

packages/channel-google/src/channel-google.ts
```diff
@@ -0,0 +1,775 @@
...
+                  asset.resource_name,
+                  asset.youtube_video_asset.youtube_video_id,
+                  asset.image_asset.file_size,
+                  asset.image_asset.full_size.url,
+                  asset.image_asset.mime_type,
+                  asset.text_asset.text
+                FROM
+                  asset
+                WHERE
+                  asset.resource_name IN (${el.adGroupAd.ad.videoResponsiveAd.videos.map((c) => `'${c.asset}'`).join(',')})
```
trixobird, line 386: you already know the adExternalId here right? If yes, we should search on the db for the ads that don't already have a creative and then fetch only these.

packages/channel-google/src/channel-google.ts
```diff
@@ -0,0 +1,775 @@
...
+                Authorization: `Bearer ${refreshedIntegration.accessToken}`,
+              },
+              body: JSON.stringify({ query }),
+            }).catch(() => {
+              throw new AError('Error fetching video ads');
+            });
+
+            const jsonResponse: unknown = await response.json();
+
+            const asset = AssetResponseSchema.parse(jsonResponse);
```
trixobird, line 404:
```suggestion
            const asset = AssetResponseSchema.safeParse(jsonResponse);
```

packages/channel-google/src/channel-google.ts
```diff
@@ -0,0 +1,775 @@
...
+      name: account.descriptiveName ?? account.resourceName,
+      currency: account.currencyCode,
+      externalId: account.id.toString(),
+    }));
+
+    return await saveAccounts(channelAccounts, integration);
+  }
+
+  async getReportStatus(_adAccount: AdAccountIntegration, _taskId: string): Promise<JobStatusEnum> {
+    return Promise.resolve(JobStatusEnum.FAILED);
```
trixobird, line 599:
```suggestion
    return Promise.reject(new AError('Not Implemented'));
```

packages/channel-google/src/channel-google.ts
```diff
@@ -0,0 +1,775 @@
...
+
+  const validatedResponse = schema.parse(googleAdsResponse);
+
+  if (!validatedResponse.results) throw new AError('No data found');
+
+  return validatedResponse;
+};
+
+const disConnectIntegrationOnError = async (integrationId: string, error: Error, notify: boolean): Promise<boolean> => {
+  const metaErrorValidatingAccessTokenChangedSession =
```
trixobird, line 759: you just copy paste it from meta it seems :)

packages/channel-google/src/channel-google.ts
```diff
@@ -0,0 +1,775 @@
...
+  accessToken: string,
+  schema: ZodSchema<GenericResponse<T>>,
+  query?: string,
+  customerId?: string,
+): Promise<GenericResponse<T> | AError> => {
+  const managerId = env.GOOGLE_CHANNEL_TEMP_CUSTOMER_ID;
+
+  const url = `https://googleads.googleapis.com/${apiVersion}/customers/${customerId ?? managerId}/googleAds:search`;
+
+  const defaultQuery = `
```
trixobird, line 722: I think in both cases you are passing a query so let's delete it and make query mandatatory. Same with customerId

packages/channel-google/src/channel-google.ts
```diff
@@ -0,0 +1,775 @@
...
+
+    return parsed.data.user_id;
+  }
+
+  getType(): IntegrationTypeEnum {
+    return IntegrationTypeEnum.GOOGLE;
+  }
+}
+
+const fetchGoogleAdsData = async <T>(
```
trixobird, line 712: I think we can unify handlePagination and fetchGoogleAdsData, there are doing the same thing.

packages/channel-google/src/channel-google.ts
```diff
@@ -0,0 +1,775 @@
...
+
+            const jsonResponse: unknown = await response.json();
+
+            const asset = AssetResponseSchema.parse(jsonResponse);
+
+            if (asset instanceof Error) return asset;
+
+            if (!asset.results) continue;
+
+            const flattenedCreatives: ChannelCreative[] = asset.results.map((c) => ({
```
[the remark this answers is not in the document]
dennis-adsviewer: _body_: ad_group_ad.ad.video_responsive_ad.descriptions [0]
_title_: video.title
_status_: we don't fetch this, but we could fetch **ad_group_ad.ad.status** for this
_callToActionType_: ad_group_ad.ad.video_responsive_ad.call_to_actions [0]
_imageUrl?_: no image url here, we could use video_id to construct a video url?

packages/channel-google/src/channel-google.ts
```diff
@@ -0,0 +1,775 @@
...
+      }
+
+      const adData = validatedResponse.results[0].adGroupAd;
+
+      const previewHTML = `
+      <iframe
+        title="${adData.ad.name ?? ''}"
+        src="${adData.ad.finalUrls?.[0] ?? ''}"
+        width="600"
+        height="400"
```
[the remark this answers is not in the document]
dennis-adsviewer: for now let's use 600x400

packages/channel-google/src/channel-google.ts
```diff
@@ -0,0 +1,775 @@
...
+    } catch (error) {
+      return new AError(`Something went wrong while getting the preview`);
+    }
+  }
+
+  getDefaultPublisher(): PublisherEnum {
+    return PublisherEnum.Facebook;
+  }
+
+  async saveAdAccounts(integration: Integration): Promise<DbAdAccount[] | AError> {
```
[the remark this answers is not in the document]
dennis-adsviewer: `query = """
    SELECT
        customer_client.client_customer,
        customer_client.descriptive_name,
        customer_client.currency_code,
        customer_client.time_zone,
        customer_client.manager,
        customer_client.status
    FROM
        customer_client
    WHERE
        customer_client.status = 'ENABLED'
"""`

Should work

packages/channel-google/src/channel-google.ts
```diff
@@ -0,0 +1,669 @@
...
+                  asset.id,
+                  asset.resource_name,
+                  asset.youtube_video_asset.youtube_video_id
+                FROM
+                  asset
+                WHERE
+                  asset.resource_name IN (${el.adGroupAd.ad.videoResponsiveAd.videos.map((c) => `'${c.asset}'`).join(',')})
+              `;
+
+            const response = await fetch(url, {
```
trixobird, line 312: can we use Google.handlePangination here? I think the url is the same...

packages/channel-google/src/channel-google.ts
trixobird: let's rename it to something more appropriate, like customerQuery

trixobird (commented):
resolved the vast majoriy of the past comments - good job.

added some new, and we have pending that:
- when i tried to ingress data, my insights were not saved
- add the call back url in the channel prod
- make sure that both credential work (we may need to add ourselved as test users)

packages/channel-google/src/channel-google.ts
```diff
@@ -0,0 +1,657 @@
...
+          externalAdAccountId: dbAccount.externalId,
+        }));
+
+        const adSets = youtubeData.map((el) => ({
+          externalId: String(el.adGroup.id),
+          name: el.adGroup.name,
+          externalCampaignId: el.campaign.id,
+        }));
+
+        const ads: ChannelAd[] = youtubeData.map((c) => ({
```
trixobird, line 292: I see that we don't save ads name. @dennis-adsviewer should we? if yes, which field?

packages/channel-google/src/channel-google.ts
```diff
@@ -0,0 +1,657 @@
...
+        const uniqueInsights = Array.from(
+          new Map(
+            insights.map((insight) => [
+              `${insight.externalAdId}-${insight.date.toISOString()}-${insight.device}-${insight.publisher}-${insight.position}`,
+              insight,
+            ]),
+          ).values(),
+        );
+
+        await saveInsightsAdsAdsSetsCampaigns(
```
trixobird, line 371:
```suggestion
        await deleteOldInsights(dbAccount.id, range.since, range.until);
        await saveInsightsAdsAdsSetsCampaigns(
```

This will solve the issue where the second time that you sync, sync fails.

packages/channel-google/src/channel-google.ts
```diff
@@ -0,0 +1,657 @@
...
+          campaignGroup,
+          new Map<string, string>(),
+          dbAccount,
+          adSets,
+          new Map<string, string>(),
+          ads,
+          new Map<string, string>(),
+          creatives,
+          uniqueInsights,
+        );
```
trixobird, line 381:
```suggestion
        );
        logger.info(
          `Finished ${range.since.toISOString()} - ${range.until.toISOString()} ad ingress for adAccountId: ${dbAccount.id}`,
        );
```

packages/channel-google/src/channel-google.ts
```diff
@@ -0,0 +1,657 @@
...
+
+            const flattenedCreatives: ChannelCreative[] = asset.map((c) => ({
+              externalAdId: el.adGroupAd.ad.id,
+              externalId: c.asset.id,
+              adAccountId: dbAccount.id,
+              name: c.asset.resourceName,
+              body: el.adGroupAd.ad.videoResponsiveAd?.descriptions?.[0].text,
+              title: el.video.title,
+              callToActionType: el.adGroupAd.ad.videoResponsiveAd?.callToActions?.[0].text,
+              imageUrl: c.asset.youtubeVideoAsset?.youtubeVideoId,
```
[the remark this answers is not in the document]
dennis-adsviewer: yeah, the pattern is : 'https://www.youtube.com/watch?v=' + c.asset.youtubeVideoAsset?.youtubeVideoId

packages/channel-google/src/channel-google.ts
dennis-adsviewer: Google doesn't have a concept of Ad Name, we can use the Video.name field for ad name

packages/channel-google/src/channel-google.ts
```diff
@@ -0,0 +1,657 @@
...
+                  asset
+                WHERE
+                  asset.resource_name IN (${el.adGroupAd.ad.videoResponsiveAd.videos.map((c) => `'${c.asset}'`).join(',')})
+              `;
+
+            const asset = await Google.handlePagination(integration, query, assetResponseSchema, dbAccount.externalId);
+
+            if (!asset || isAError(asset)) return;
+
+            const flattenedCreatives: ChannelCreative[] = asset.map((c) => ({
```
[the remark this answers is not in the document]
aaryansinha16: Yes, I tried to fetch status, got this error"Unrecognized field in the query: 'ad_group_ad.ad.status'."
Tried to look for some other keys similar to this, but couldn't find any. 

@dennis-adsviewer

packages/channel-google/src/channel-google.ts
```diff
@@ -0,0 +1,657 @@
...
+      body: params.toString(),
+    }).catch(() => {
+      throw new AError(`Failed to refresh token`);
+    });
+
+    if (response instanceof Error) return response;
+
+    const tokenData: unknown = await response.json();
+
+    const schema = z.object({
```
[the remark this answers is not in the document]
aaryansinha16:
```typescript
    const tokenData: unknown = await response.json();

    const schema = z.object({
      access_token: z.string().optional(),
      expires_in: z.number().int().optional(),
      scope: z.string().optional(),
      error: z.string().optional()
    });

    const parsed = schema.safeParse(tokenData);
    if (!parsed.success) {
      logger.error(parsed.error, 'Failed to parse refresh access token response');
      return new AError('Failed to parse refresh access token response');
    }

    if(parsed.data.access_token && parsed.data.expires_in){
      return await updateIntegrationTokens(integration, {
        accessToken: parsed.data.access_token,
        accessTokenExpiresAt: addInterval(new Date(), 'seconds', parsed.data.expires_in),
        refreshToken: integration.refreshToken,
        refreshTokenExpiresAt: undefined,
      });
    }

    if (parsed.data.error) {
      const error = new Error(parsed.data.error)
      if(await disConnectIntegrationOnError(integration.id, error, true)) return new AError('Invalid refresh token')
    }

    return new AError('Failed to refresh access token');
``` 

This works, but do you think this can have a much better version programmatically?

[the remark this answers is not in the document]
trixobird: I made a commit. let me know if it is ok

packages/channel-google/src/channel-google.ts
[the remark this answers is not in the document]
dennis-adsviewer: Try:
ad_group_ad.status

trixobird (approved):

PR #634, opened 2024-10-24
524 improved onboarding flow

elevchyt:
The easiest way to view this feature is to create a new user.


apps/web/src/middleware.ts
```diff
@@ -129,6 +131,17 @@ export async function middleware(request: NextRequest): Promise<NextResponse> {
     const redirectUrl = new URL(DEFAULT_HOME_PATH, request.url);
     return NextResponse.redirect(redirectUrl);
   }
+
+  // If user has the "Onboarding" milestone, redirect them to the onboarding page
+  if (
+    tokenData &&
+    tokenData.milestones.includes(Milestones.Onboarding) &&
```
trixobird, line 138:
```suggestion
     // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- existing jwts will not have this field
    tokenData.milestones?.includes(Milestones.Onboarding) &&
```

apps/web/src/app/(authenticated)/introduction/page.tsx
trixobird, line 1: I was expecting a bigger aray for the video. Also the buttons label are hidden
image

if I refresh it looks ok

apps/web/src/app/(authenticated)/actions.ts
```diff
@@ -21,3 +23,27 @@ export const refreshJWTToken = async (): Promise<RefreshTokenQuery> => await urq
...
+): Promise<UrqlResult> => {
+  try {
+    const res = await handleUrqlRequest(urqlClientSdk().removeUserMilestone(values));
+    if (!res.success) {
+      return {
+        success: false,
+        error: res.error,
+      };
+    }
+    await changeJWT(res.data.removeUserMilestone.token, res.data.removeUserMilestone.refreshToken);
```
trixobird, line 38: in the changeJWT we use cookies() multiple times. We can put them in a variable for slightly faster actions `const cookiesHandle = cookies()`

apps/web/src/app/(authenticated)/introduction/components/video-card.tsx
```diff
@@ -0,0 +1,61 @@
...
+          component={Link}
+          href="/integrations"
+          target="_blank"
+          leftSection={<IconPlugConnected />}
+          variant="gradient"
+          fullWidth
+          mt="md"
+          radius="md"
+        >
+          Go to Integrations
```
trixobird, line 44: translations? applicable in line 56 as well.

apps/web/src/app/(authenticated)/introduction/layout.tsx
```diff
@@ -0,0 +1,15 @@
+import React, { Suspense } from 'react';
+import { Title, Text } from '@mantine/core';
+import { useTranslations } from 'next-intl';
+import LoaderCentered from '@/components/misc/loader-centered';
+
+export default function Layout({ children }: { children: React.ReactNode }): React.ReactNode {
```
trixobird, line 6:
```suggestion
export default function IntroductionLayout({ children }: { children: React.ReactNode }): React.ReactNode {
```

apps/web/src/app/(authenticated)/introduction/page.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { Milestones } from '@/graphql/generated/schema-server';
+import wavingPeopleAnimation from '../../../../public/lotties/waving-people.json';
+import { removeUserMilestoneAndGetJWT } from '../actions';
+import VideoCard from './components/video-card';
+
+export default function Introduction(): React.ReactNode {
+  useEffect(() => {
+    // On page load, remove the onboarding milestone from the user
+    // so that they can navigate the rest of the app
+    void removeUserMilestoneAndGetJWT({
```
trixobird, line 16: this is mostly for @dennis-adsviewer 
the moment that the user is visiting the introduction page, we are removing them from the `new-user` group. Which means that next time they will be directed to the usual top ads page. I was thinking that if a users does not have an integration they should be directed here?

dennis-adsviewer: Correct, until an integration is added they should keep seeing this page.

apps/web/src/app/(authenticated)/introduction/page.tsx
elevchyt: It's because the video hasn't properly loaded, I'm looking into it.

elevchyt: Probably fixed, curious to see if it happens on production as well.

apps/web/src/middleware.ts
trixobird: This is still an issue. If I have an old jwt without millestones and change to this branch, web breaks. We need to put this code just for the release, and then we can remove it.

apps/web/src/middleware.ts
trixobird, line 1: I now noticed that if I visit http://localhost:3000/ I am redirected to introduction, but if I visit the summary (http://localhost:3000/summary?fetchPreviews=true&chartMetric=impressions&orderBy=impressions_abs&dateFrom=1729285200000&dateTo=1729890000000) then I am not.
This is also happening when you login (I tried with social)

elevchyt: Hmm so we should only limit the allowed pages to be visited to be integrations and organization (aka the ones that the two video buttons point to)?

trixobird (approved):

PR #648, opened 2024-10-29
fix to avoid double re-renders of top ads

elevchyt:
was an easy fix

### DIFF
--- a/apps/web/src/app/(authenticated)/summary/components/top-ads-container.tsx
+++ b/apps/web/src/app/(authenticated)/summary/components/top-ads-container.tsx
@@ -3,7 +3,7 @@
 import { Title, Flex, Select, Text, type ComboboxItem } from '@mantine/core';
 import { useAtom, useAtomValue } from 'jotai';
 import { usePathname, useRouter, useSearchParams } from 'next/navigation';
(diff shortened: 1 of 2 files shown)


trixobird (commented):
I put a `logger.info(args.filter.groupBy, "nikos")` in line 52 if `insights-operations.ts` and it was triggered 29 times. 2 times with publisher only and 27 for the topads. It seems that are rendered 3 times (I can confirm it visually as well)

elevchyt:
> I put a `logger.info(args.filter.groupBy, "nikos")` in line 52 if `insights-operations.ts` and it was triggered 29 times. 2 times with publisher only and 27 for the topads. It seems that are rendered 3 times (I can confirm it visually as well)

Hmm this makes sense since we're making a request for each of the 9 publishers, so assuming you tested it using strict mode 9 * 2 = 18 which means there is an extra request "batch" happening for no reason.

elevchyt:
Found the culprit! Turns out the re-rendering was happening due to the initial userDetails data and the loaded data flag in the main-shell.tsx having an async problem with the assignment of their values.
Fixed and looking to be working as expected :)

trixobird (approved):

PR #668, opened 2024-11-12
608 user level creative feedbackcomments

elevchyt:
Functionality for adding user comments to creatives.

_User mentions will be done in another issue because most libraries are broken or don't support React 19._


packages/utils/src/shared-schemas.ts
```diff
@@ -10,3 +10,7 @@ export const sendFeedbackSchema = z.object({
   type: z.string(),
   message: messageSchema,
 });
+
+export const commentSchema = z.object({
```
trixobird, line 14: should we also use it in the comment-operations.ts as well?

trixobird (commented):
I am not able to create a comment in a creative, where is the btn? :)

elevchyt:
> I am not able to create a comment in a creative, where is the btn? :)

You need to set group by "Creative ID" and remove anything else to make it clearer.
Top right of the insight card.

I'll make creative ID the default grouping like we discussed.

trixobird (commented):
- how do I delete/edit comment?
- I think creatives with comments should have a different icon, what do you think?

May not for this issue, or any issue
- Should we move from comments to conversations(threaded) like slack?
- Should we add resolve/re-open functionality in conversations?
@dennis-adsviewer would that make sense or it is enough for now?

elevchyt:
> * I think creatives with comments should have a different icon, what do you think?
I also wanted to do that, but we'll have to fetch comments for every single creative which would perform multiple requests at once.
Right now, we are fetching each creative's comments when the user actually opens the comments of a creative.

> * how do I delete/edit comment?
You can't delete/edit, I can create the functionality if we believe it's going to be needed.

> * I think creatives with comments should have a different icon, what do you think?
Yeah, I'll do that.

trixobird (approved):

PR #671, opened 2024-11-14
matcher warning fix

elevchyt:
only removes the match warning that floods the terminal


apps/web/src/middleware.ts
```diff
@@ -31,10 +31,6 @@ const publicPaths = [
   '/api/auth/reset-password',
 ];
 
-export const config = {
```
trixobird, line 34: I think this is applicable in backoffice as well no?

trixobird (approved):

PR #694, opened 2024-11-24
655: introduce threshold in insights in backend

trixobird:


elevchyt (approved):
Nice, I'll do the FE soon :)

PR #704, opened 2024-11-29
703: Fetch ad previews when group by creativeId

trixobird:


apps/server/src/schema/integrations/insights-operations.ts
```diff
@@ -262,16 +262,19 @@ const GroupedInsightDto = builder.simpleObject(
       type: IFrameDTO,
       nullable: true,
       resolve: async (root, _args, _ctx, _info) => {
-        if (!root.adId) return null;
+        if (!root.adId && !root.creativeId) return null;
```
elevchyt, line 265: I haven't tested it, but shouldn't it be || instead of &&?

apps/web/src/app/(authenticated)/insights/components/order-filters.tsx
```diff
@@ -162,13 +162,21 @@ export default function OrderFilters(): React.ReactNode {
           label={t('adPreviewsTooltip')}
           refProp="rootRef"
           position="top-start"
-          disabled={!isPending && isParamInSearchParams(searchParams, urlKeys.groupedBy, InsightsColumnsGroupBy.adId)}
+          disabled={
+            !isPending &&
```
elevchyt, line 166:
```suggestion
            !isPending ||
```

apps/server/src/schema/integrations/insights-operations.ts
elevchyt, line 265:
```suggestion
        if (!root.adId || !root.creativeId) return null;
```
I haven't tested it, but shouldn't it be || instead of &&?

elevchyt (approved):

PR #707, opened 2024-11-30
706: spend is divided 2 times by 100

trixobird:
Leftery we divide by `/ 100` in one more place, but it is the spendEur so I believe it is fine


elevchyt (approved):
Looks fine :)

PR #718, opened 2024-12-07
717 groupby position yield wrong results

trixobird:
## Description

Lefteri as part of this issue, I fixed a bug on the web. I noticed that if you select dates and then copy paste the url, the date range was -1. This was because mantine datepicker does not take into account timezone. So it was setting hours to 0, while being gmt +2, which is the previsou day 22:00 in gmt.


elevchyt (approved):
Looks good to me 👍

Issue #711, opened 2024-12-04
Select grid size (number of previews per row)


dennis-adsviewer:
Priority 2 (low) 

We have received requests to make the amount of previews per row flexible, user defined - some users have large monitors and would prefer maybe 4,5 or 6 per row vs. 3 which is forced now.

elevchyt:
Sounds like a nice feature, I'll look into it.

PR #720, opened 2024-12-09
get and update insights per row user preferences

elevchyt:
Adds preferences per user.

At the moment, there is only an `insightsPerRow` preference which does what https://github.com/adsviewer/turboviewer/issues/711 says.


apps/server/src/schema/user/preferences/preferences-operations.ts
```diff
@@ -0,0 +1,24 @@
...
+
+builder.mutationFields((t) => ({
+  updatePreferences: t.withAuth({ isInOrg: true }).prismaField({
+    type: PreferencesDto,
+    args: {
+      idToUpdate: t.arg.string({ required: true }),
+      insightsPerRow: t.arg.int({ required: false }),
+    },
+    resolve: async (query, _parent, args, _ctx) => {
+      const dataToBeUpdated = {
```
trixobird, line 13: I would inline this :)

apps/server/src/schema/user/user-types.ts
```diff
@@ -102,6 +103,7 @@ export const UserDto = builder.prismaObject('User', {
     currentOrganization: t.relation('currentOrganization', { nullable: true, type: OrganizationDto }),
     comments: t.relation('comments', { nullable: false }),
     taggedInComment: t.relation('taggedInComment', { nullable: false }),
+    preferences: t.relation('preferences', { nullable: true, type: PreferencesDto }),
```
trixobird, line 106:
```suggestion
    preferences: t.relation('preferences', { nullable: true }),
```
pothos is smart and knows

apps/server/src/schema/user/user-types.ts
```diff
@@ -12,6 +12,7 @@ import { type InputShapeFromFields } from '@pothos/core';
 import { builder } from '../builder';
 import { OrganizationDto } from '../organization/org-types';
 import { type GraphQLContext } from '../../context';
+import { PreferencesDto } from './preferences/preferences-types';
```
trixobird, line 15:
```suggestion
```

apps/web/src/app/(authenticated)/insights/components/order-filters.tsx
```diff
@@ -52,6 +56,11 @@ export default function OrderFilters(): React.ReactNode {
     return '12';
   };
 
+  const getInsightsPerRowValue = (): string => {
+    if (userDetails.preferences) return String(userDetails.preferences.insightsPerRow);
+    return '3';
+  };
+
```
trixobird, line 63:
```suggestion
  const getInsightsPerRowValue = (): string =>
    userDetails.preferences ? String(userDetails.preferences.insightsPerRow) : '3';

```

apps/web/src/app/(authenticated)/insights/components/order-filters.tsx
```diff
@@ -76,39 +85,59 @@ export default function OrderFilters(): React.ReactNode {
...
+  const handlePageSizeChange = (_value: string | null, option: ComboboxItem): void => {
     resetInsights();
     const newURL = addOrReplaceURLParams(pathname, searchParams, urlKeys.pageSize, option.value);
     startTransition(() => {
       router.replace(newURL);
     });
   };
 
-  const handleOrderByChange = (value: string | null, option: ComboboxItem): void => {
+  const hanleInsightsPerRowChange = (_value: string | null, option: ComboboxItem): void => {
```
trixobird, line 104:
```suggestion
  const handleInsightsPerRowChange = (_value: string | null, option: ComboboxItem): void => {
```

apps/web/src/components/insights/insights-grid.tsx
```diff
@@ -31,7 +34,10 @@ export default function InsightsGrid(props: PropsType): ReactNode {
       ) : null}
 
       {/* Render insights */}
-      <SimpleGrid cols={{ base: 1, sm: 2, lg: 3 }} style={{ display: 'relative' }}>
+      <SimpleGrid
+        cols={{ base: 1, sm: 2, lg: userDetails.preferences?.insightsPerRow ?? 3 }}
```
trixobird, line 38: this is the second time that we are hard coding 3. I know that this is not needed, since preferences will always be there, but maybe we can put it as a variable somewhere.

apps/web/src/app/(authenticated)/insights/components/order-filters.tsx
```diff
@@ -202,6 +231,22 @@ export default function OrderFilters(): React.ReactNode {
           />
         </Flex>
 
+        {/* Insights per row setting */}
+        <Flex align="flex-end" mr="sm">
+          <Select
+            description={t('insightsPerRow')}
```
trixobird, line 237: image
allignment...

apps/web/src/app/(authenticated)/insights/components/order-filters.tsx
```diff
@@ -202,6 +231,22 @@ export default function OrderFilters(): React.ReactNode {
           />
         </Flex>
 
+        {/* Insights per row setting */}
+        <Flex align="flex-end" mr="sm">
```
trixobird, line 235: I think we don't need this options on mobile view.

apps/web/src/app/(authenticated)/insights/components/order-filters.tsx
trixobird: image
some else broke :P

trixobird (approved):

apps/web/src/app/(authenticated)/insights/components/order-filters.tsx
elevchyt: wtf

elevchyt: refresh or do pnpm dev again, it's something in your local

PR #727, opened 2024-12-11
default date range in analytical view

elevchyt:


apps/web/src/app/(authenticated)/insights/components/order-filters.tsx
```diff
@@ -275,6 +275,7 @@ export default function OrderFilters(): React.ReactNode {
                 disabled={isPending}
               />
               <DatePickerInput
+                description={tGeneric('pickDateRange')}
```
trixobird, line 278: I don't think that we need this. The idea is that there will be no case where this DatePicker is not filled.

trixobird (commented):
hmm does not work, maybe I didn't describe it accurately? We want to have the datepicker prefilled with the default the same way that we are doing in summary.
image

